PR #2, opened 2018-09-20
Add tnozicka

tnozicka:
/cc @deads2k


tnozicka:
merging manually to unblock other efforts - in line with what you have approved yesterday for the other repo

PR #4, opened 2018-09-21
Fix scripts

tnozicka:


deads2k:
Is this going to allow us to remove the CI bit where we are doing go get before the script call?

tnozicka:
not this one in particular but if we vendor those we can use the same pattern and avoid `go get` at all

this one is just a fix for newer go that don't need GOPATH set so it's not exported by default and the script fails

deads2k:
> not this one in particular but if we vendor those we can use the same pattern and avoid `go get` at all

I don't want to vendor things that are not code dependencies, but I am ok `go get`-ing things.

PR #10, opened 2018-10-01
switch to daemonset

deads2k:
Also fixes config merging.

/assign @mfojtik


mfojtik:
@deads2k we also need to switch to hostNetwork and locking if we want to get the cluster up working , i will send a follow up PR's

PR #15, opened 2018-10-03
ensure operator config exists on start

mfojtik:


deads2k:
/lgtm
/hold 
holding because michal said so.

PR #18, opened 2018-10-10
add image references

juanvallejo:
cc @deads2k

### DIFF
--- a/manifests/image-references
+++ b/manifests/image-references
@@ -1,0 +1,12 @@
+kind: ImageStream
+apiVersion: image.openshift.io/v1
+spec:
+  tags:
+  - name: cluster-openshift-controller-manager-operator
+    from:
+      kind: DockerImage
+      name: docker.io/openshift/origin-cluster-openshift-controller-manager-operator:v4.0
+  - name: hypershift
+    from:
+      kind: DockerImage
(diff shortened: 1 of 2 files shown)


manifests/image-references
```diff
@@ -0,0 +1,12 @@
+kind: ImageStream
+apiVersion: image.openshift.io/v1
+spec:
+  tags:
+  - name: cluster-openshift-controller-manager-operator
+    from:
+      kind: DockerImage
+      name: docker.io/openshift/origin-cluster-openshift-controller-manager-operator:latest
```
sttts: latest? Don't we version them?

sttts: Do we still push to docker.io?

juanvallejo: Thanks, done

juanvallejo: I will do this for the other three operators as well

deads2k:
@tnozicka something looks stuck here with travis.  

manually merging

deads2k:
interesting.  it self-unstuck itself after I mucked with settings.

tnozicka:
@deads2k sorry I missed this, cleaning the inbox now.

Why do we still have Travis? I don't have perms but I though last time we agreed to disable it on all of our repos - that's why I haven't made it explicitly optional yet.

PR #20, opened 2018-10-15
report status to the CVO

deads2k:
Uses https://github.com/openshift/cluster-version-operator/pull/31, a dynamic client, and the operator status used by several operator to create a control loop that will keep the `ClusterOperator` synchronized if the API is available.

/cc @abhinavdahiya 
/assign @mfojtik @sttts @sanchezl 

/hold

hold until we roughly agree on the upstream shape.  This will require us to re-fit some conditions.

### DIFF
--- a/vendor/github.com/openshift/library-go/pkg/operator/status/controller.go
+++ b/vendor/github.com/openshift/library-go/pkg/operator/status/controller.go
@@ -1,0 +1,206 @@
+package status
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/golang/glog"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+
+	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+	"github.com/openshift/library-go/pkg/operator/v1alpha1helpers"
+)
+
+var workQueueKey = "instance"
+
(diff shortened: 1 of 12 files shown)


deads2k:
@tnozicka any idea what is up with Travis and tide?

vendor/github.com/openshift/library-go/pkg/operator/status/controller.go
```diff
@@ -0,0 +1,199 @@
...
+func (c StatusSyncer) sync() error {
+	currentDetailedStatus, err := c.operatorStatusProvider.CurrentStatus()
+	if apierrors.IsNotFound(err) {
+		return c.clusterOperatorClient.Delete(c.clusterOperatorName, nil)
+	}
+	if err != nil {
+		return err
+	}
+
+	operatorConfig, err := c.clusterOperatorClient.Get(c.clusterOperatorName, metav1.GetOptions{})
```
sttts: odd the get the same object again.

vendor/github.com/openshift/library-go/pkg/operator/status/controller.go
```diff
@@ -0,0 +1,199 @@
...
+		if apierrors.IsNotFound(createErr) {
+			// this means that the API isn't present.  We did not fail.  Try again later
+			c.queue.AddRateLimited(workQueueKey)
+			return nil
+		}
+		if createErr != nil {
+			return createErr
+		}
+	}
+	if updateErr != nil {
```
sttts: nit: return updateErr

sttts:
some nits

vendor/github.com/openshift/library-go/pkg/operator/status/controller.go
deads2k: > odd the get the same object again.

Uncommon, but dynamic informers/listers are in 1.13 and this is fairly slow moving.

vendor/github.com/openshift/library-go/pkg/operator/status/controller.go
```diff
@@ -0,0 +1,199 @@
...
+
+	return nil
+}
+
+func OperatorConditionToClusterOperatorCondition(condition *operatorv1alpha1.OperatorCondition) (map[string]interface{}, error) {
+	if condition == nil {
+		return nil, nil
+	}
+	buf := &bytes.Buffer{}
+	if err := json.NewEncoder(buf).Encode(condition); err != nil {
```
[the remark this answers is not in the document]
deads2k: > ugly. I guess we don't have go types yet of this object?

For the CVO ClusterOperator status?  Not yet.

PR #25, opened 2018-10-18
Use available helper function to do config check

damemi:
Use helper function now available in library-go, and update deps


damemi:
cc @deads2k

PR #26, opened 2018-10-29
observe internal registry hostname

bparees:
I'm not sure if any controllers *directly* consume this value today, but the image_change_controller is directly impacted when it changes because of the way it caches imagestream values (and the imagestream values are affected by this value due to the storage decorator that uses this value). 

So ensuring that the controller manager restarts if/when this value changes, is necessary to ensure the image_change_controller is operating on the correct values.

We also might want to update the service account controller to consume this value in the future.

### DIFF
--- a/pkg/operator/observe_config.go
+++ b/pkg/operator/observe_config.go
@@ -10,8 +10,6 @@ import (
 	"github.com/golang/glog"
 
 	"k8s.io/client-go/informers"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/client-go/util/workqueue"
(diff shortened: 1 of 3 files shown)


bparees:
@deads2k @juanvallejo ptal

/cc @adambkaplan

bparees:
(this is basically a cut+paste of what @juanvallejo did in https://github.com/openshift/cluster-openshift-apiserver-operator/pull/41)

juanvallejo:
This lgtm

bparees:
@deads2k bump

pkg/operator/observe_config.go
```diff
@@ -120,6 +127,28 @@ func observeControllerManagerImagesConfig(kubeClient kubernetes.Interface, confi
...
+	configImage, err := imageConfigClient.Images().Get("cluster", metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		return observedConfig, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+	internalRegistryHostName := configImage.Status.InternalRegistryHostname
+	if len(internalRegistryHostName) > 0 {
+		unstructured.SetNestedField(observedConfig, internalRegistryHostName, "imagePolicyConfig", "internalRegistryHostname")
```
adambkaplan: @bparees iirc the observedConfig is an `OpenshiftControllerManagerConfig` object, so the copy-paste from @juanvallejo won't fully translate here.

bparees: I would expect controllermanagerconfig and api config to contain the same configuration information (an equivalent to the master-config.yaml that both of them consumed pre-4.0).

But @deads2k can weigh in

deads2k: it is a struct of this information: https://github.com/openshift/api/blob/master/openshiftcontrolplane/v1/types.go#L141

the API types are different because the binaries are different and require different information.

deads2k:
@bparees where does it cache the value from?  If it's discovering it from the cluster, I'd suggest removing that discovery and driving it via a configuration value.  Then this becomes a natural thing to do.

Also, I think you've found the correct place to wire this in.

bparees:
> @bparees where does it cache the value from? 

It caches them from the apiserver.  It lists the imagestreams (which come back decorated based on the internalRegistryHostname value as the apiserver knows it).  So if it does that list *before* the internalRegistryHostname is set, the values it caches are, basically, wrong, until a relist occurs.

> If it's discovering it from the cluster, I'd suggest removing that discovery and driving it via a configuration value. Then this becomes a natural thing to do.

Not really feasible.  The image change controller would become responsible for mutating the imagestreams it retrieves based on the internalRegistryHostname, which means it's then potentially disconnected from the values that the api itself is actually returning to everyone else.  We wouldn't recommend/require every client of the imagestream api do its own internalRegistryHostname lookup + replacement, so I don't see why we'd do it here.

> Also, I think you've found the correct place to wire this in.

sounds like I need to add an internalRegistryHostname field to your OpenShiftControllerManagerConfig struct though.  Or, since all I really want to do is trigger a rollout, find another way to trigger the rollout.

bparees:
https://github.com/openshift/api/pull/119 adds a field to the controllermanagerconfig where we can store this value (we're going to need it eventually anyway), assuming that gets approved/merged i'll come back and update this PR to store the value there.

pkg/operator/observe_config.go
bparees: ok i've fixed this, but holding pending creation of tests.

bparees:
@deads2k tested this locally, seems to work, but since it's futzing w/ your logic still, i'd like your final sign off.

@adambkaplan hopefully this is really close so you can build on top of it.

PR #27, opened 2018-10-31
Observe Cluster BuildController config

adambkaplan:
Add capability to observe the config.Build CRD, and update the
build controller's configuration appropriately.
This does not update the additional trusted CA used by the build
controller.

JIRA-ID: [DEVEXP-154](https://jira.coreos.com/browse/DEVEXP-154)


pkg/operator/observe_config.go
```diff
@@ -149,6 +150,63 @@ func observeInternalRegistryHostname(kubeClient kubernetes.Interface, clientConf
...
+	if len(nodeSelector.MatchLabels) > 0 {
+		unstructured.SetNestedField(observedConfig, nodeSelector.MatchLabels, "build", "buildOverrides", "nodeSelector")
+	}
+	tolerations := build.Spec.BuildOverrides.Tolerations
+	if len(tolerations) > 0 {
+		unstructured.SetNestedField(observedConfig, tolerations, "build", "buildOverrides", "tolerations")
+	}
+	// TODO: 1) generate CA bundle from ConfigMapRef
+	//       2) write CA bundle to a ConfigMap in the controller-manager's namespace
+	//       3) wire in logic to force a rollout if CA bundle changes
```
bparees: // (4) add a watch to watch the CA bundle configmap(pointed to by the ConfigMapRef) so that when it changes, the operator can update the bundle in the controller-manager's configmap and do a rollout.

bparees: (and of course if the configmapref changes, the watch needs to be replaced/updated to watch the new referenced configmap)

I believe @deads2k has some utilities/patterns for doing a single resource watch that he can point you to.

manifests/0000_21_cluster-openshift-controller-manager-operator_01_config.crd.yaml
```diff
@@ -13,4 +13,18 @@ spec:
     categories:
     - coreoperators
   subresources:
-    status: {}
\ No newline at end of file
+    status: {}
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
```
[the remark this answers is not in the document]
deads2k: > This operator is responsible for creating the `config.Build` CRD.

This is not the correct location.  We want as little as possible managed by the CVO because it significantly constrains future choice.  If you need to have the operator create a CRD, let the operator make a CRD.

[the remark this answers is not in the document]
deads2k: /cc @openshift/sig-master in case it comes up in other operators.  We don't want to do this.

pkg/operator/observe_config.go
```diff
@@ -149,6 +150,63 @@ func observeInternalRegistryHostname(kubeClient kubernetes.Interface, clientConf
 	return observedConfig, nil
 }
 
+func observeBuildControllerConfig(kubeClient kubernetes.Interface, clientConfig *rest.Config, observedConfig map[string]interface{}) (map[string]interface{}, error) {
```
juanvallejo: You'll need to add this function to the list of observeConfigFuncs in the constructor

manifests/0000_21_cluster-openshift-controller-manager-operator_01_config.crd.yaml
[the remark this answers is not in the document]
bparees: @deads2k please elaborate on how this constrains future choice so we can understand the argument you are making.  From my perspective this is defining a resource name/group, nothing more, and a future operator could delete it entirely if that were deemed necessary, so writing code to do it instead of dropping a yaml file in a directory is unnecessary work against an unclear future risk.


/cc @derekwaynecarr @jwforres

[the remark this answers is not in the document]
deads2k: https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go#L32-L43 is a good spot to do it.

The CVO removes items that are no longer in the payload.  Doing it our sync loop allows ownership to change in the future.  Putting it in the CVO payload pins it for all time (or until someone builds a special migrator)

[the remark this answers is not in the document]
bparees: @deads2k and I spoke in his cube... the way forward for this is:

1) add the CRD to openshift/api, ala: https://github.com/openshift/api/pull/118
2) vendor the .yaml in from openshift/api to this repo
3) add logic in the operator that installs the yaml, ala https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go#L32-L44  (not sure this is precisely where this particular CRD would go since it is not v311 related though, probably need more specific insight from @deads2k about where he wants to see it).

adambkaplan:
@bparees ptal

pkg/operator/observe_config.go
```diff
@@ -149,13 +154,172 @@ func observeInternalRegistryHostname(listers Listers, observedConfig map[string]
...
+		return nil
+	}
+	var err error
+	switch v := val.(type) {
+	case int64, bool:
+		err = unstructured.SetNestedField(observedConfig, v, nestedFields...)
+	case string:
+		if len(v) > 0 {
+			err = unstructured.SetNestedField(observedConfig, v, nestedFields...)
+		}
```
bparees: i'm not sure how i feel about baking in the assumption that we'd never want/need to set a field to empty string, but I don't think I'm a fan.  I think i prefer that the caller makes the decision about whether to set the field or not, in that case.

pkg/operator/observe_config.go
```diff
@@ -149,13 +154,172 @@ func observeInternalRegistryHostname(listers Listers, observedConfig map[string]
...
+			unstructured.SetNestedMap(observedConfig, v, nestedFields...)
+		}
+	default:
+		err = fmt.Errorf("unsupported field type %T", v)
+	}
+	return err
+}
+
+// envVarsToRawEnvVars converts a `corev1.EnvVar` slice to a JSON-serializable `interface{}` slice.
+func envVarsToRawEnvVars(val []corev1.EnvVar) ([]interface{}, error) {
```
bparees: so I still need to actually managed to test it, but I did this and at least in my unit tests it appeared to work:

```
	t, err := Convert(configImage.Spec.AllowedRegistriesForImport)
	if err != nil {
		return observedConfig, err
	}
	unstructured.SetNestedField(observedConfig, t, "imagePolicyConfig", "allowedRegistriesForImport")
```

```
func Convert(o interface{}) (interface{}, error) {
	if o == nil {
		return nil, nil
	}
	buf := &bytes.Buffer{}
	if err := json.NewEncoder(buf).Encode(o); err != nil {
		return nil, err
	}
	//ret := map[string]interface{}{}
	ret := []interface{}{}
	if err := json.NewDecoder(buf).Decode(&ret); err != nil {
		return nil, err
	}

	return ret, nil
}
```

I still need to look at the actual rawextension yaml it's producing to confirm it, but superficially I guess I would expect it to just do the right thing.

pkg/operator/observe_config.go
adambkaplan: You can make the same argument for all types. Is it worth keeping this function by adding a parameter, or a separate function def?

bparees: I intended to make the argument for all types, sorry that wasn't clear :)

I don't have a strong preference between adding an "skipEmpty" bool and making this function always invoke the set and relying on the caller to not call it if they don't want it set.

I think i have a slight preference towards the latter in case the decision making gets even more subtle (e.g. different behavior for a nil pointer to a slice vs an empty slice)

pkg/operator/observe_config.go
bparees: i was finally able to test this and shockingly it did exactly what i wanted.

```
apiVersion: config.openshift.io/v1
kind: Image
metadata:
  annotations:
    imageregistry.operator.openshift.io/checksum: sha256:90cd0cb4759af7f0d936ff489f7baf1f64186b2943ea7f35bea5e4e32afe726e
  creationTimestamp: 2018-11-13T23:30:29Z
  generation: 1
  name: cluster
  namespace: ""
  resourceVersion: "14857"
  selfLink: /apis/config.openshift.io/v1/images/cluster
  uid: 1b1c9421-e79c-11e8-93d1-9acfc1405147
spec:
  additionalTrustedCA:
    name: ""
    namespace: ""
  allowedRegistriesForImport:
  - domainName: foo3.bar.com
  - domainName: foo2.bar.com
status:
  internalRegistryHostname: image-registry.openshift-image-registry.svc.cluster.local:5000
```

was observed into:
```
apiVersion: openshiftapiserver.operator.openshift.io/v1alpha1
kind: OpenShiftAPIServerOperatorConfig
metadata:
  creationTimestamp: 2018-11-13T23:24:06Z
  generation: 5
  name: instance
  resourceVersion: "15491"
  selfLink: /apis/openshiftapiserver.operator.openshift.io/v1alpha1/openshiftapiserveroperatorconfigs/instance
  uid: 3700d195-e79b-11e8-91d9-ee096c38eb37
spec:
  imagePullPolicy: ""
  imagePullSpec: registry.svc.ci.openshift.org/openshift/origin-v4.0-20181113211753@sha256:d8c99f7fc22bd8e387ef5cbf85d69c4f1ed59bb17da0944beb33a333d963add7
  logging:
    level: 4
    vmodule: ""
  managementState: Managed
  observedConfig:
    imagePolicyConfig:
      allowedRegistriesForImport:
      - domainName: foo3.bar.com
      - domainName: foo2.bar.com
      internalRegistryHostname: image-registry.openshift-image-registry.svc.cluster.local:5000
```

adambkaplan:
@bparees ptal

pkg/operator/observe_config.go
```diff
@@ -149,13 +153,126 @@ func observeInternalRegistryHostname(listers Listers, observedConfig map[string]
...
+	if err = observeField(observedConfig, build.Spec.BuildOverrides.ImageLabels, "build.buildOverrides.imageLabels", true); err != nil {
+		return nil, fmt.Errorf("failed to observe %s: %v", "build.buildOverrides.imageLabels", err)
+	}
+	nodeSelector := build.Spec.BuildOverrides.NodeSelector
+	if err = observeField(observedConfig, nodeSelector.MatchLabels, "build.buildOverrides.nodeSelector", true); err != nil {
+		return nil, fmt.Errorf("failed to observe %s: %v", "build.buildOverrides.nodeSelector", err)
+	}
+	// Control plane config does not support MatchExpressions yet
+	if len(nodeSelector.MatchExpressions) > 0 {
+		return nil, fmt.Errorf("config.Build: %s is not supported", "buildOverrides.nodeSelector.matchExpressions")
```
bparees: i'd be inclined to log this rather than error out entirely.

adambkaplan:
@bparees updated and squashed

bparees:
@adambkaplan have you been able to test this on a cluster to confirm the cluster config gets reflected into the controller-manager's config in the way you expect?

adambkaplan:
@bparees I have been able to run a full e2e test locally, and it works -ish:

1. The observed config from the build CRD shows up as expected in the operator config:
```
apiVersion: openshiftcontrollermanager.operator.openshift.io/v1alpha1
kind: OpenShiftControllerManagerOperatorConfig
metadata:
  name: instance
spec:
...
  observedConfig:
    build:
      buildDefaults:
        env:
        - name: FOO
          value: BAR
        - name: RED
          value: HAT
        gitHTTPProxy: http://user:password@myserver.proxy.org
        gitHTTPSProxy: https://user:password@myserver.proxy.org
        imageLabels:
        - name: config.openshift.io
          value: v1
        - name: build.config.openshift.io
          value: v2
      buildOverrides:
        imageLabels:
        - name: config.openshift.io
          value: v1
        - name: build.config.openshift.io
          value: v2
        tolerations: null
...
```
2. The observed config is copied to the controller manager, but it's serializing in JSON instead of YAML
```
apiVersion: v1
data:
  config.yaml: |
    {"apiVersion":"openshiftcontrolplane.config.openshift.io/v1","build":{"buildDefaults":{"env":[{"name":"FOO","value":"BAR"},{"name":"RED","value":"HAT"}],"gitHTTPProxy":"http://user:password@myserver.proxy.org","gitHTTPSProxy":"https://user:password@myserver.proxy.org","imageLabels":[{"name":"config.openshift.io","value":"v1"},{"name":"build.config.openshift.io","value":"v2"}]},"buildOverrides":{"imageLabels":[{"name":"config.openshift.io","value":"v1"},{"name":"build.config.openshift.io","value":"v2"}],"tolerations":null},"imageTemplateFormat":{"format":"registry.svc.ci.openshift.org/openshift/origin-v4.0-20181115075802@sha256:a6dde9d8810aa6c46e081f6527b30eaf27be25e2780cdf4faf8513188a732b15"}},"deployer":{"imageTemplateFormat":{"format":"registry.svc.ci.openshift.org/openshift/origin-v4.0-20181115075802@sha256:54115ce20cfc3f1a6883efd8c150c37ff166bdbb50c13b13ea590d2133c48bbf"}},"kind":"OpenShiftControllerManagerConfig"}
kind: ConfigMap
```

Also not clear why `tolerations` is serializing, when it should be skipped.

adambkaplan:
@deads2k just change the controller config filename to `config.json` so we manage admin/dev expectations?

PR #28, opened 2018-11-02
Add Build CRD to Manifests

adambkaplan:
Update the operator to sync the CRD spec for `config.Build` (not the actual cluster instance - just the CRD definition).


pkg/operator/resource/apiextensions/resourceapply/crd.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
+)
+
+var (
+	apiExtCodecs = scheme.Codecs.UniversalDeserializer()
+)
+
+// ApplyCustomResourceDefinitionV1Beta1 merges object metadata and CustomResourceDefinition spec.
+func ApplyCustomResourceDefinitionV1Beta1(client clientv1beta1.CustomResourceDefinitionsGetter, required *v1beta1.CustomResourceDefinition) (*v1beta1.CustomResourceDefinition, bool, error) {
+	existing, err := client.CustomResourceDefinitions().Get(required.Name, metav1.GetOptions{})
```
adambkaplan: This is a candidate for `library-go`

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -42,6 +45,9 @@ func syncOpenShiftControllerManager_v311_00_to_latest(c OpenShiftControllerManag
 		"v3.11.0/openshift-controller-manager/sa.yaml",
 		"v3.11.0/openshift-controller-manager/svc.yaml",
 	)
+	vendorResourceResults := apiextapply.ApplyDirectly(c.apiExtensionClient, vendor_assets.Asset,
+		"vendor/github.com/openshift/api/config/v1/build.crd.yaml")
+	directResourceResults = append(directResourceResults, vendorResourceResults...)
```
adambkaplan: CRD definition applied here in the sync loop.

bparees:
@smarterclayton imho this is a ridiculous amount of code to replace "drop buildcrd yaml file into manifests dir, let the CVO sort it out".

Do we really think this is necessary?  If so, why?

smarterclayton:
I don’t see value in syncing CRDs in the operator vs having cvo sync it unless you are trying to be clever with CRDs which is probably bad.  

For instance, I recently removed cleverness in CVO that merges CRDs.  CVO reconciles CRDs every few minutes

adambkaplan:
@bparees updated, assuming CVO will take care of the rest.

bparees:
/lgtm

Thanks, please go back and remove it from openshift/api so there's no confusion.

adambkaplan:
@bparees the YAML for the conifg.Build CRD was never merged to master in `api`. I closed the pull which added it (see comment ref above).

bparees:
ah, cool.

PR #30, opened 2018-11-07
bump(*)

adambkaplan:
Pulling in API fixes

/assign @bparees 
/cc @deads2k 

This seems like an enormous number of new deps...


bparees:
[shrug]
if it builds+passes tests i'm not going to worry about it right now.

/lgtm

PR #31, opened 2018-11-07
use pullalways for the operator image

bparees:


bparees:
@deads2k kube api operator is doing this, openshiftapi operator is not.  Personally I think pull always is preferred, at a minimum it makes development easier when you're iterating on pushing new versions of the image to the same tag and restarting the pod.

And i don't think it really has any meaningful cost.

/assign @deads2k

PR #32, opened 2018-11-08
Manage Build Controller Additional Trusted CA

adambkaplan:
* Controller manager mounts additional CA bundle from ConfigMap
* Operator watches build CRD and configmaps in openshift-config for changes
* Operator validates additional CA data, merges into a single PEM block
* Observed additional CA recorded in operator config spec, including sha1 hash
* Primary sync loop merges additional CA data into a bundle mounted and read by the openshift-controller-manager


pkg/operator/observe_config.go
```diff
@@ -156,6 +159,26 @@ func observeInternalRegistryHostname(listers Listers, observedConfig map[string]
 	return observedConfig, nil
 }
 
+// observeBuildAdditionalCAPath observes the data in the controller's generated
+func observeBuildAdditionalCAIsSet(listers Listers, observedConfig map[string]interface{}) (map[string]interface{}, error) {
+	// observed state of the build controller's Additional CA bundle is stored in a ConfigMap within the operator namespace
+	// this observes if there is data in this ConfigMap, and if the ca bundle exists
+	caMap, err := listers.configmapLister.ConfigMaps(operatorNamespaceName).Get("openshift-build-additional-ca")
```
adambkaplan: Note - the "observed CA" ConfigMap will be added in a later PR.

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -143,3 +149,15 @@ func manageOpenShiftControllerManagerPublicConfigMap_v311_00_to_latest(client co
 
 	return resourceapply.ApplyConfigMap(client, configMap)
 }
+
+func manageBuildAdditionalCAConfigMap(client coreclientv1.ConfigMapsGetter) (*corev1.ConfigMap, bool, error) {
+	caMap := resourceread.ReadConfigMapV1OrDie(v311_00_assets.MustAsset("v3.11.0/openshift-controller-manager/build-additional-ca-cm.yaml"))
```
adambkaplan: This syncs the observed CA bundle into the controller-manager's additional-ca configMap.

bparees:
(I also likely have an insufficient understanding of the existing operator behavior, especially around how it handles mounted resources)

adambkaplan:
> Are we now going to have 3 copies of the configmap?
> 
> 1. the configmap the user defined in some namespace and pointed to via ConfigMapReference
> 2. the copy of (1) that the operator puts in its own namespace to represented the observed/desired content to apply to the controller
> 3. the copy of (2) that the operator puts into the controller's namespace so the controller can actually mount/consume it?

Yes, but (2) is not an exact copy. The operator will read the first user-provided configmap, merge the content (verifying proper encoding in the process), and then copy into the second configmap as the canonical "observed" CA. Hence why in this PR we are watching the configmaps in the core operators namespace.

This approach gives us two main advantages:

1. Admins can easily see the merged CA bundle that the operator wants to apply to the controller manager
2. Changes to the canonical observed CA trigger rollout.

However, I agree the additional configMap is a big minus and may not be necessary. Worth exploring storing hashes in the operator status.

adambkaplan:
(I also say _will_ because I plan to add the logic to watch the build CRD and associated ConfigMap in a separate PR)

pkg/apis/openshiftcontrollermanager/v1alpha1/types.go
```diff
@@ -39,10 +39,24 @@ type OpenShiftControllerManagerOperatorConfigSpec struct {
 	// observedConfig holds a sparse config that controller has observed from the cluster state.  It exists in spec because
 	// it causes action for the operator
 	ObservedConfig runtime.RawExtension `json:"observedConfig"`
+
+	// additionalTrustedCA references the additional trusted certificate authorities that operator should attempt to configure for the build controller.
+	AdditionalTrustedCA *AdditionalTrustedCA `json:"additionalTrustedCA,omitempty"`
```
adambkaplan: This breaks the "no pointers" rule for API, but results in much cleaner code.
This will likely go away when certs are split out by hostname.

@deads2k do we need protobuf as well as json?

pkg/apis/openshiftcontrollermanager/v1alpha1/types.go
```diff
@@ -39,10 +39,24 @@ type OpenShiftControllerManagerOperatorConfigSpec struct {
...
+}
+
+type AdditionalTrustedCA struct {
+	// sha1Hash contains the sha1 hash of the CA bundle data
+	SHA1Hash string `json:"sha1Hash,omitempty" protobuf:"bytes,1,opt,name=sha1Hash"`
+
+	// configMapName is the name of the ConfigMap in the openshift-config namespace containing the additional trusted CAs for the build controller.
+	ConfigMapName string `json:"configMap,omitempty" protobuf:"bytes,2,opt,name=configMap"`
```
adambkaplan: CA data stores the SHA1Hash of the merged CA bundle, and the name of the ConfigMap.
Per discussion with @bparees and @deads2k, cluster admin provided configurations will live in the `openshift-config` namespace.

pkg/operator/observe_config.go
```diff
@@ -72,23 +77,27 @@ func NewConfigObserver(
...
+			observeBuildAdditionalCA,
 		},
 		listers: Listers{
-			imageConfigLister: configInformer.Config().V1().Images().Lister(),
-			configmapLister:   kubeInformersForOpenshiftCoreOperators.Core().V1().ConfigMaps().Lister(),
-			buildConfigLister: configInformer.Config().V1().Builds().Lister(),
+			imageConfigLister:            configInformer.Config().V1().Images().Lister(),
+			coreOperatorsConfigMapLister: kubeInformersForOpenshiftCoreOperators.Core().V1().ConfigMaps().Lister(),
+			buildConfigLister:            configInformer.Config().V1().Builds().Lister(),
+			clusterConfigConfigMapLister: openshiftConfigInformer.Core().V1().ConfigMaps().Lister(),
```
adambkaplan: Lister for ConfigMaps in the `openshift-config` namespace.

pkg/operator/observe_config.go
```diff
@@ -72,23 +77,27 @@ func NewConfigObserver(
...
 		},
 	}
 
 	c.operatorConfigSynced = operatorConfigInformer.Informer().HasSynced
-	c.configmapSynced = kubeInformersForOpenshiftCoreOperators.Core().V1().ConfigMaps().Informer().HasSynced
+	c.coreOperatorsConfigSynced = kubeInformersForOpenshiftCoreOperators.Core().V1().ConfigMaps().Informer().HasSynced
```
adambkaplan: Do we still need to watch the `openshift-core-operators` namespace?

pkg/operator/observe_config.go
```diff
@@ -97,28 +106,29 @@ func (c ConfigObserver) sync() error {
 	var err error
 	observedConfig := map[string]interface{}{}
 
+	operatorConfig, err := c.operatorConfigClient.Get("instance", metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	currentCASpec := operatorConfig.Spec.AdditionalTrustedCA.DeepCopy()
```
adambkaplan: Fetch operator config first, which can be mutated by the `observer` functions.
Deep copy the current CA spec so we can exit early if no updates are needed.

pkg/operator/observe_config.go
```diff
@@ -292,6 +294,50 @@ func ConvertJSON(o interface{}) (interface{}, error) {
...
+		operatorConfig.Spec.AdditionalTrustedCA = nil
+		return observedConfig, nil
+	}
+	cm, err := listers.clusterConfigConfigMapLister.ConfigMaps(openshiftConfigNamespaceName).Get(build.Spec.AdditionalTrustedCA.Name)
+	if errors.IsNotFound(err) {
+		operatorConfig.Spec.AdditionalTrustedCA = nil
+		glog.Warningf("referenced configMap %s/%s for additional trusted certificate authorities does not exist",
+			openshiftConfigNamespaceName,
+			build.Spec.AdditionalTrustedCA.Name)
+		return observedConfig, nil
```
adambkaplan: If build CRD references a ConfigMap that doesn't exist, log the warning and treat it as if no CA had been defined.

pkg/operator/observe_config.go
```diff
@@ -292,6 +294,50 @@ func ConvertJSON(o interface{}) (interface{}, error) {
...
+	}
+	h := sha1.New()
+	h.Write(caData)
+	caHash := fmt.Sprintf("%x", h.Sum(nil))
+	operatorConfig.Spec.AdditionalTrustedCA = &operatorconfigv1alpha1.AdditionalTrustedCA{
+		SHA1Hash:      caHash,
+		ConfigMapName: cm.Name,
+	}
+	unstructured.SetNestedField(observedConfig, "/var/run/configmaps/additional-ca/additional-ca.crt", "build", "additionalTrustedCA")
```
adambkaplan: Additional CA's for the build controller need two pieces of config in `Spec`:

1. The hash and reference to the ConfigMap with the CA data
2. The mount location for the build controller's configuration.

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -107,11 +119,16 @@ func manageOpenShiftControllerManagerConfigMap_v311_00_to_latest(client coreclie
...
-func manageOpenShiftControllerManagerDeployment_v311_00_to_latest(client appsclientv1.DaemonSetsGetter, options *v1alpha1.OpenShiftControllerManagerOperatorConfig, previousAvailability *operatorsv1alpha1.VersionAvailability, forceRollout bool) (*appsv1.DaemonSet, bool, error) {
+func manageOpenShiftControllerManagerDeployment_v311_00_to_latest(client appsclientv1.DaemonSetsGetter, options *v1alpha1.OpenShiftControllerManagerOperatorConfig, previousAvailability *operatorsv1alpha1.VersionAvailability, additionalCAHash string, forceRollout bool) (*appsv1.DaemonSet, bool, error) {
 	required := resourceread.ReadDaemonSetV1OrDie(v311_00_assets.MustAsset("v3.11.0/openshift-controller-manager/ds.yaml"))
 	required.Spec.Template.Spec.Containers[0].Image = options.Spec.ImagePullSpec
 	required.Spec.Template.Spec.Containers[0].Args = append(required.Spec.Template.Spec.Containers[0].Args, fmt.Sprintf("-v=%d", options.Spec.Logging.Level))
 
+	if required.Annotations == nil {
+		required.Annotations = make(map[string]string)
+	}
+	required.Annotations["config.openshift.io/build.additional-ca-hash"] = additionalCAHash
```
adambkaplan: Hash of the additional CA is added as an annotation to the controller manager daemonset, ensuring rollout on update.

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -143,3 +160,78 @@ func manageOpenShiftControllerManagerPublicConfigMap_v311_00_to_latest(client co
...
+	caMap := resourceread.ReadConfigMapV1OrDie(v311_00_assets.MustAsset("v3.11.0/openshift-controller-manager/build-additional-ca-cm.yaml"))
+	buildConfig, err := configClient.Builds().Get("cluster", metav1.GetOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return "", false, err
+	}
+	if buildConfig == nil {
+		caMap.Data = nil
+		_, modified, err := resourceapply.ApplyConfigMap(client, caMap)
+		operatorConfig.Status.AdditionalTrustedCA = nil
+		return "", modified, err
```
adambkaplan: If no build CRD, clear the data in the controller manager's ConfigMap and nil the status for `AdditionalTrustedCA`

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -143,3 +160,78 @@ func manageOpenShiftControllerManagerPublicConfigMap_v311_00_to_latest(client co
...
+
+	caData, err := mergeCAConfigMap(client, buildConfig.Spec.AdditionalTrustedCA)
+	if err != nil {
+		return "", false, err
+	}
+	if len(caData) == 0 && operatorConfig.Status.AdditionalTrustedCA != nil {
+		caMap.Data = nil
+		_, modified, err := resourceapply.ApplyConfigMap(client, caMap)
+		operatorConfig.Status.AdditionalTrustedCA = nil
+		return "", modified, err
```
adambkaplan: Handle case of empty data in referenced ConfigMap

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -143,3 +160,78 @@ func manageOpenShiftControllerManagerPublicConfigMap_v311_00_to_latest(client co
...
+	}
+	h := sha1.New()
+	h.Write(caData)
+	caHash := fmt.Sprintf("%x", h.Sum(nil))
+	if operatorConfig.Status.AdditionalTrustedCA != nil &&
+		operatorConfig.Status.AdditionalTrustedCA.SHA1Hash == caHash {
+		return caHash, false, nil
+	}
```
adambkaplan: If hashes are equivalent, no updates needed.

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -143,3 +160,78 @@ func manageOpenShiftControllerManagerPublicConfigMap_v311_00_to_latest(client co
...
+	operatorConfig.Status.AdditionalTrustedCA = &v1alpha1.AdditionalTrustedCA{
+		SHA1Hash:      caHash,
+		ConfigMapName: buildConfig.Spec.AdditionalTrustedCA.Name,
+	}
+	caMap.Data["additional-ca.crt"] = string(caData)
+	_, modified, err := resourceapply.ApplyConfigMap(client, caMap)
+	return caHash, modified, err
```
adambkaplan: If updates are happening, copy the merged data into the ConfigMap, which is then mounted into the build controller

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -143,3 +160,78 @@ func manageOpenShiftControllerManagerPublicConfigMap_v311_00_to_latest(client co
...
+			if block == nil && len(rest) > 0 {
+				return nil, fmt.Errorf("map key %s contains invalid PEM data: %s", key, string(rest))
+			}
+			err := pem.Encode(pemData, block)
+			if err != nil {
+				return nil, err
+			}
+			data = rest
+		}
+	}
```
adambkaplan: mergePEMData ensures that the provided map contains valid PEM-encoded data for every key. This code may live on even if we are not merging CAs, as IMO the operator should validate that the provided certificates have valid encoding.

pkg/apis/openshiftcontrollermanager/v1alpha1/types.go
bparees: SHA probably belongs only in status, not spec, as it is not something the user gets to set.

adambkaplan: I did this for consistency, though we can probably get away without it.

bparees: if we don't need it, let's get rid of it.  just a source of confusion otherwise.

adambkaplan: Now I remember...I do this so we can capture that the cluster admin updated the data in the referenced ConfigMap. I want to avoid races due to having an informer for `openshift-config` here and in the main operator controller.

adambkaplan:
Closing due to change in direction, but keeping branch alive for other operators that need to accomplish similar things.

adambkaplan:
/hold

Will never merge, but leaving this up for teams with similar requirements.

adambkaplan:
This functionality was added in https://github.com/openshift/origin/pull/21614

PR #36, opened 2018-11-21
add e2e test for cluster-manager-operator

bparees:


test/e2e/main_test.go
```diff
@@ -0,0 +1,73 @@
...
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+	if err != nil {
+		return err
+	}
+	kclient, err := k8sclient.NewForConfig(config)
+	if err != nil {
+		return err
+	}
+
+	err = wait.PollImmediate(1*time.Second, 10*time.Minute, func() (bool, error) {
```
soltysh: @bparees any particular reason you're not using the same pattern as in https://github.com/openshift/cluster-image-registry-operator/blob/1500c6432360d2eb0328ab4e537ab0c3b9214207/test/e2e/main_test.go#L41-L61 ?

bparees: if you're asking about the client, it's because that uses the operator SDK for the client and this repo doesn't use the operator SDK.

soltysh: Shouldn't we be consistent across the board and have similar patterns?

bparees: @soltysh and someday maybe we will.  Today we have operators written in many different ways, vendoring many different libraries, so the tests reflect that.

PR #43, opened 2018-11-30
operator: wait for spec.observedConfig to be fill before installing

sttts:


deads2k:
/hold

I suspect that @bparees would prefer not to do this.

bparees:
> I suspect that @bparees would prefer not to do this.

it seems like a hack(to avoid a controller manager restart i assume?), but is there another reason i would prefer not to do it?

sttts:
> it seems like a hack(to avoid a controller manager restart i assume?), but is there another reason i would prefer not to do it?

Why should we install static pods before looking at cluster configuration the installer has put there? I pretty strongly think this would be wrong and asks for inconsistent seconds/minutes until another static pod is installed.

Note: this is not about later runs. It is about cluster config (e.g. network CIDRs) that is put into the cluster early for a reason.

bparees:
@sttts @deads2k is this still at all relevant/needed?

coreydaley:
@sttts @deads2k bump

PR #47, opened 2018-12-18
Git Proxy Support

adambkaplan:
Observe git proxy config settings from updated config API.


pkg/operator/observe_config.go
```diff
@@ -172,23 +172,39 @@ func observeBuildControllerConfig(listers Listers, observedConfig map[string]int
...
+		})
+		buildEnv = append(buildEnv, corev1.EnvVar{
+			Name:  "HTTPS_PROXY",
+			Value: build.Spec.BuildDefaults.DefaultProxy.HTTPSProxy,
+		})
+		buildEnv = append(buildEnv, corev1.EnvVar{
+			Name:  "NO_PROXY",
+			Value: build.Spec.BuildDefaults.DefaultProxy.NoProxy,
+		})
+	}
```
bparees: this needs to be handled differently.  builddefaults.env is for env variables that are supposed to be added to the application image that is built.  it is not for env vars to be added to the build pod (more importantly, it's likely undesirable to do this because these vars will end up in the application image).

please limit this PR to fixing up the gitproxy defaulting part for now as the correct solution to the defaultproxy is going to require more pieces.

adambkaplan: @bparees removed.

PR #50, opened 2019-01-09
fix progressing condition logic

deads2k:
progressing is based on comparing the daemonset generation to itself and the operator generation to itself.  

@bparees flake fix
@sanchezl I think I found a better spot for the operator status generation ptal

### DIFF
--- a/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
+++ b/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
@@ -2,6 +2,7 @@ package operator
 
 import (
 	"fmt"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


bparees:
thanks @deads2k

deads2k:
removed fancy set of observed generation because it probably caused
```
"message": "daemonset/controller-manager: observed generation is 44, desired generation is 45.\nopenshiftcontrollermanageroperatorconfigs/instance: observed generation is 0, desired generation is 1.",
```

vrutkovs:
This fixes BYOR install at least.

/retest

deads2k:
merged to unstick us in the 3pm hour

sttts:
:+1:

PR #52, opened 2019-01-10
Tighten config api

adambkaplan:
Shorten top-level API to align with other renames.


bparees:
@adambkaplan assuming this gets merged can you followup w/ a rename of the v1alpha1 api to v1?

adambkaplan:
bump @bparees or @deads2k for approval

manifests/0000_21_cluster-openshift-controller-manager-operator_01_config.crd.yaml
```diff
@@ -1,15 +1,16 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  name: openshiftcontrollermanageroperatorconfigs.openshiftcontrollermanager.operator.openshift.io
+  name: configs.openshiftcontrollermanager.operator.openshift.io
```
bparees: i'm waiting to hear from @deads2k on this because he hasn't done this rename in the kube controller operator.

personally I prefer this rename, but i also prefer to keep this operator in sync w/ the kube controller operator and that has a higher precedence for me.

adambkaplan:
/hold

Need to resolve conflict with #55

manifests/0000_21_cluster-openshift-controller-manager-operator_01_config.crd.yaml
bparees: ok @deads2k and i spoke and he has another set of thoughts on what the long term name of these apis should be, so let's put a hold on this PR for now.

his vision is something more like:  openshiftcontrollermanagers(resoure) and  operator.openshift.io (group).

upside:
you can say "oc edit openshiftcontrollermanagers" w/o collisions w/ other resource names (don't need to fully qualiy it).

downside, if the controllermanager operator ever needs a second resource type, it doesn't have its own obvious group to add, so you would have to add sometihng like "openshiftcontrollermanagerfoo".

Personally I prefer each operator have its own api group for that reason (future proofing for adding more resource types easily) and just a general isolation, but it does have a bit of a downside on the cli experience.

@deads2k also had some reasons for preferring his approach for if/when these apis move into openshift/api, but i'm not sure that's a requirement either.

tldr: discussions need to happen.

bparees:
i think this is obsolete.  we've moved the config api into openshift/api and done the rename there. i'll be following up w/ the updates to this repo to pick up those apis once https://github.com/openshift/client-go/pull/81 lands.

@adambkaplan feel free to reopen if i've missed something.

PR #54, opened 2019-01-11
remove references to operator v1alpha1 API

adambkaplan:
Looks like this was missed in https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/42


bparees:
/lgtm

but this also needs to be renamed/cleaned up: https://github.com/openshift/cluster-openshift-controller-manager-operator/tree/master/pkg/apis/openshiftcontrollermanager/v1alpha1

PR #56, opened 2019-01-14
update to new build nodeselector override api

bparees:


bparees:
/assign @adambkaplan 

@adambkaplan ptal

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -143,7 +144,7 @@ func syncOpenShiftControllerManager_v311_00_to_latest(c OpenShiftControllerManag
 
 func manageOpenShiftAPIServerClientCA_v311_00_to_latest(client coreclientv1.CoreV1Interface, recorder events.Recorder) (bool, error) {
 	const apiserverClientCA = "client-ca"
-	_, caChanged, err := resourceapply.SyncConfigMap(client, recorder, kubeAPIServerNamespaceName, apiserverClientCA, targetNamespaceName, apiserverClientCA)
+	_, caChanged, err := resourceapply.SyncConfigMap(client, recorder, kubeAPIServerNamespaceName, apiserverClientCA, targetNamespaceName, apiserverClientCA, []metav1.OwnerReference{})
```
adambkaplan: would like @deads2k to weigh in on what the owner ref is used for.

bparees: From what I saw it replaces the ownerrefs that may exist on the source configmap, with the ownerrefs you pass here.  So in this case the target CM gets no ownerrefs.

it's not obvious to me what, if anything, the owner should be, and the other operators haven't even been updated to the new api yet, so i punted.

PR #59, opened 2019-01-18
update operator status when the operand is unmanaged

bparees:
https://bugzilla.redhat.com/show_bug.cgi?id=1667376

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -7,6 +7,7 @@ import (
 	"github.com/golang/glog"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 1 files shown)


bparees:
> @bparees do we want all our team's operators to follow this pattern wrt Unmanaged and the three clusteroperator status conditions ?

@gabemontero it's as reasonable as anything else i've come up with...  the main point of debate was whether available should be true/false(depending on the operand) or unknown.  It is possible(at least in some cases) for an operator to understand if its operand is available or not, even when it's not actively managing that operand.  I chose not to implement that logic at least for now (ie pending pushback).

pkg/operator/operator.go
```diff
@@ -80,10 +83,36 @@ func (c OpenShiftControllerManagerOperator) sync() error {
...
 	switch operatorConfig.Spec.ManagementState {
-	case operatorsv1.Unmanaged:
+	case operatorv1.Unmanaged:
+		// manage status
+		originalOperatorConfig := operatorConfig.DeepCopy()
+		v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorv1.OperatorCondition{
+			Type:    operatorv1.OperatorStatusTypeAvailable,
+			Status:  operatorv1.ConditionUnknown,
+			Reason:  "Unmanaged",
+			Message: "the controller manager is in an unmanaged state, therefore its availability is unknown.",
```
coreydaley: s/its/it's

pkg/operator/operator.go
```diff
@@ -80,10 +83,36 @@ func (c OpenShiftControllerManagerOperator) sync() error {
...
+		})
+		v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorv1.OperatorCondition{
+			Type:    operatorv1.OperatorStatusTypeProgressing,
+			Status:  operatorv1.ConditionFalse,
```
coreydaley: Should this also be unknown instead of false?

pkg/operator/operator.go
```diff
@@ -80,10 +83,36 @@ func (c OpenShiftControllerManagerOperator) sync() error {
...
+			Reason:  "Unmanaged",
+			Message: "the controller manager is in an unmanaged state, therefore no changes are being applied.",
+		})
+		v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorv1.OperatorCondition{
+			Type:    operatorv1.OperatorStatusTypeFailing,
+			Status:  operatorv1.ConditionFalse,
```
coreydaley: Should this also be unknown?

PR #60, opened 2019-01-19
add metric scraping to controller

bparees:
https://bugzilla.redhat.com/show_bug.cgi?id=1666118


bparees:
@coreydaley @gabemontero ptal.  follows roughly the same pattern as https://github.com/openshift/cluster-image-registry-operator/pull/165 except that this operator manages its resources slightly differently (some are managed by the CVO like our other operators, but some are actually managed by the operator itself, somewhat like the registry operator).

I had to move the controller-manager namespace+service monitor resources to be managed by the CVO because:
1) the apply logic the operator uses to manage its own resources doesn't understand the ServiceMonitor api (and it would be a significant library-go change to fix that)
2) the namespace has to exist to create the service monitor

/assign @coreydaley @gabemontero

bparees:
@coreydaley @gabemontero rebased and ready for review again

OWNERS
```diff
@@ -4,6 +4,9 @@ reviewers:
   - sttts
   - mfojtik
   - soltysh
+  - adambkaplan
+  - coreydaley
```
coreydaley: I don't know if I'm ready for this kind of responsibility ...

bparees: interestingly the bot let you lgtm https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/59 anyway, so apparently this doesn't even matter....

gabemontero:
@bparees - the v3.11 bindata .... my first time look at this controller/operator says
a) there is a version of this controller in 3.11in the `openshift-controller-manager` namespace (which I also see in my 4.0 cluster
b) so are there elements of the 3.11 metrics from that controller that we are re-using?
c) or is this some sort of migration play (which I thought we did not have to worry about)? ... or are we talking about migrating "apps" like whatever current monitoring customers may have?

From the 4.0 side, seems like a complete set of certs, sa's, endpoints, http/https services that I would expect for establishing a metrics endpoint to be scraped ... though I've never set up one of those, so aside from me getting a general sense, in case I need to replicate this somewhere, I get the gist ... if there are specific wrong with it, I'm unaware 

I guess I'll wait on any responses to my question in case they remind you of something additionally needed before applying a LGTM.  Also gives @coreydaley more time to chime in more than he has already if he wants

bparees:
> @bparees - the v3.11 bindata .... my first time look at this controller/operator says
> a) there is a version of this controller in 3.11in the openshift-controller-manager namespace (which I 
> also see in my 4.0 cluster

yes, the controller itself existed in 3.11 and had metrics being scraped.  I would ignore the naming of the bindata dir though, I think @deads2k named it that way when the intent was to upgrade from 3.11 to "latest" (4.0) but now it basically just represents the initial starting resources for the controller operand that is created by the controller operator.


> b) so are there elements of the 3.11 metrics from that controller that we are re-using?

for the controller it's the same metrics as existed in 3.11..  this PR does two things:
1) configures prometheus to be aware of the controller process so it can scrape the metrics that the controller was already making available (as it did in 3.11)
2) exposes the metrics the operator publishes(defines a service for them) and configures prometheus to scrape that service to collect the operator metrics


> c) or is this some sort of migration play (which I thought we did not have to worry about)? ... or are we > talking about migrating "apps" like whatever current monitoring customers may have?

no migration here, and nothing to do w/ customer apps/metrics.  


> From the 4.0 side, seems like a complete set of certs, sa's, endpoints, http/https services that I would 
> expect for establishing a metrics endpoint to be scraped ... though I've never set up one of those, so 
> aside from me getting a general sense, in case I need to replicate this somewhere, I get the gist ... if 
> there are specific wrong with it, I'm unaware

The controller process already sets up certs via the service signing controller:

https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/bindata/v3.11.0/openshift-controller-manager/svc.yaml#L7
https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/bindata/v3.11.0/openshift-controller-manager/ds.yaml#L51
https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/bindata/v3.11.0/openshift-controller-manager/ds.yaml#L42


(the actual logic for wiring the metrics up to the /metrics path is part of the controller process and again, already existed, carried forward from 3.11).

So with respect to the controller process, the only change is to configure prometheus to be aware of/scrape that existed endpoint/service.

The service that exposes the operator metrics is defined here(introduced by this PR), note again the use of service signing certs:
https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/60/files#diff-6f066d2e20aa9f5cb6f9c98774eb0c9bR5

The operator deployment mounts the cert here:
https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/manifests/0000_63_openshift-controller-manager-operator_07_deployment.yaml#L50

There actually was one piece of that missing which this PR corrects (ensuring the secret created by the service signing controller gets mounted into the operator pod): 
https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/60/files#diff-e11cea844d238adda4ede4b0bacf925eR39

Again the rest of the operator code already handled serving metrics on the /metrics path, using the certs (once they were correctly mounted into the pod), so the remaining task was to configure prometheus to scrape the new operator metrics service.

PR #61, opened 2019-01-21
move openshift controller manager later in CVO payload

deads2k:
nothing earlier than 70 should require openshift apis


bparees:
/hold

@deads2k i did a more significant renaming here along w/ some new files, so i'd rather just merge that, unless you see issues w/ the metrics enablement i did there:
https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/62

deads2k:
/hold cancel

this is related to fixing out queue.  If it does green, it needs to merge

PR #64, opened 2019-01-22
refactor operator to match siblings

bparees:


bparees:
/assign @adambkaplan 

@adambkaplan i think you have the most experience working in this operator, ptal.

Note that there are definitely potential followups around emitting events when observing config changes, and adding objectrefs to aid in mustgather, but this at least gets us looking like the other operators and keeps us from getting further behind if more config observation code gets added.

pkg/operator/configobservation/builds/observe_builds.go
```diff
@@ -0,0 +1,127 @@
...
+		return prevObservedConfig, append(errs, err)
+	}
+	if len(currentGitNoProxy) > 0 {
+		err := unstructured.SetNestedField(prevObservedConfig, currentGitNoProxy, gitNoProxyPath...)
+		if err != nil {
+			return prevObservedConfig, append(errs, err)
+		}
+	}
+
+	if !listers.BuildConfigSynced() {
```
adambkaplan: @deads2k we don't wait for cache sync any more?

pkg/operator/configobservation/interfaces.go
```diff
@@ -0,0 +1,27 @@
...
+	BuildConfigLister  configlistersv1.BuildLister
+	ConfigMapLister    corelistersv1.ConfigMapLister
+	ImageConfigSynced  cache.InformerSynced
+	BuildConfigSynced  cache.InformerSynced
+	ConfigMapSynced    cache.InformerSynced
+	PreRunCachesSynced []cache.InformerSynced
+}
+
+func (l Listers) ResourceSyncer() resourcesynccontroller.ResourceSyncer {
+	return nil
```
adambkaplan: looks like openshift/cluster-openshift-apiserver-operator has non-nil returns, which drives the cache sync.
See apiserver [interfaces.go](https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/pkg/operator/configobservation/interfaces.go) and [observe_config_controller.go](https://github.com/openshift/cluster-openshift-apiserver-operator/blob/8aa4775bcff1c0e3d62606006cef70e4fda3d8f9/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go#L46)

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -0,0 +1,50 @@
...
+		ConfigObserver: configobserver.NewConfigObserver(
+			operatorClient,
+			eventRecorder,
+			configobservation.Listers{
+				ImageConfigLister: configInformers.Config().V1().Images().Lister(),
+				ImageConfigSynced: configInformers.Config().V1().Images().Informer().HasSynced,
+				BuildConfigLister: configInformers.Config().V1().Builds().Lister(),
+				BuildConfigSynced: configInformers.Config().V1().Builds().Informer().HasSynced,
+				ConfigMapLister:   kubeInformersForOperatorNamespace.Core().V1().ConfigMaps().Lister(),
+				ConfigMapSynced:   kubeInformersForOperatorNamespace.Core().V1().ConfigMaps().Informer().HasSynced,
```
adambkaplan: two items:
1. Pass in a `ResourceSyncer`?
2. Pre-run cache sync for the operator config object?

See [cluster-openshift-apiserver-operator](https://github.com/openshift/cluster-openshift-apiserver-operator/blob/8aa4775bcff1c0e3d62606006cef70e4fda3d8f9/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go#L46)

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -0,0 +1,50 @@
...
+			},
+			images.ObserveInternalRegistryHostname,
+			builds.ObserveBuildControllerConfig,
+			deployimages.ObserveControllerManagerImagesConfig,
+		),
+	}
+
+	kubeInformersForOperatorNamespace.Core().V1().ConfigMaps().Informer().AddEventHandler(c.EventHandler())
+	configInformers.Config().V1().Images().Informer().AddEventHandler(c.EventHandler())
```
adambkaplan: Need an event handler for `Builds`, too.

pkg/operator/configobservation/interfaces.go
bparees: this isn't about cache sync.  resourcesyncer copies resources between namespaces.  this operator does not use it anywhere in the code path.

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
bparees: 1) we don't have a resourcesyncer to pass because we don't have any resources we are syncing across namepaces.

2) i didn't see an obvious use/value in the pre cache sync given that we need to check it in the observers.

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
bparees: well that is interesting, because the existing code didn't have one either, so i'm not sure how this was working.....

adambkaplan: sounds like we need an integration test. Given that we'll need to change cluster config to drive the test, not sure we have a clear path to build these tests at present.

bparees: it can be done in our e2e suite.

bparees:
@adambkaplan added precache syncing though i'm not really sure what purpose it serves.

also added e2e tests to confirm we are observing build+image cluster config objects properly.

still don't think we have any use for resourcesyncer so i have left that out.

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
adambkaplan: https://github.com/openshift/origin/pull/21864 starts us down that path. Each test adds 1-2 minutes to the total suite.

bparees: I actually meant it can be added in the e2e suite on this repo, that's what i ended up doing.  (the test that sets the cluster config and ensures it propagates to the operand config).

adambkaplan: yes, but I think this gets us halfway there. IMO a full e2e should:

1. Modify the cluster config
2. Verify the operator processed the config change
3. Verify the config change rolls out to the operand
4. Verifies the expected behavior changed (ex: with `BuildDefaults`)

I'd love to test steps 3 and 4 in this repo, but atm origin has better test frameworks on this front.

PR #65, opened 2019-01-28
bump to k8s 1.12 deps

bparees:


bparees:
@mfojtik fyi, thanks for the heads up

PR #66, opened 2019-01-30
use openshift/api+client-go for operator config types

bparees:


pkg/operator/generic_client.go
```diff
@@ -3,64 +3,64 @@ package operator
...
+		return operatorapiv1.OperatorStatus{}, err
 	}
 
 	return instance.Status.OperatorStatus, nil
 }
 
-func (c *operatorClient) GetOperatorState() (*operatorv1.OperatorSpec, *operatorv1.OperatorStatus, string, error) {
-	instance, err := c.informers.Openshiftcontrollermanager().V1().OpenShiftControllerManagerOperatorConfigs().Lister().Get("instance")
+func (c *operatorClient) GetOperatorState() (*operatorapiv1.OperatorSpec, *operatorapiv1.OperatorStatus, string, error) {
+	instance, err := c.informers.Operator().V1().OpenShiftControllerManagers().Lister().Get("instance")
```
adambkaplan: possibly worth addressing in a separate PR - rename instance to `cluster`?

bparees: i haven't kept up w/ what the other operators have done, but personally i like instance.

oc edit openshiftcontrollermanager.operator.openshift.io/instance makes more sense to me than oc edit openshiftcontrollermanager.operator.openshift.io/cluster

we used cluster for true "config-only" data, but in this case you're really driving an instance of something.  (the only instance of it that exists).

bparees:
/hold

I think there might be rbac changes needed.


Ben Parees | OpenShift

bparees:
/hold cancel

think i found the last spot that needed to be updated to point to the new resource name.

PR #70, opened 2019-02-07
no need for post-startup cache sync checks

sanchezl:
No need for post-startup cache sync checks for CRDs now managed by **cluster-config-operator**.

### DIFF
--- a/pkg/operator/configobservation/images/observe_images.go
+++ b/pkg/operator/configobservation/images/observe_images.go
@@ -31,11 +31,6 @@ func ObserveInternalRegistryHostname(genericListers configobserver.Listers, reco
 		}
 	}
 
-	if !listers.ImageConfigSynced() {
-		glog.Warning("images.config.openshift.io not synced")
-		return prevObservedConfig, errs
-	}
-
 	// now gather the cluster config and turn it into the observed config
 	observedConfig := map[string]interface{}{}
 	configImage, err := listers.ImageConfigLister.Get("cluster")
--- a/pkg/operator/configobservation/interfaces.go
+++ b/pkg/operator/configobservation/interfaces.go
@@ -12,9 +12,6 @@ type Listers struct {
 	ImageConfigLister  configlistersv1.ImageLister
 	BuildConfigLister  configlistersv1.BuildLister
 	ConfigMapLister    corelistersv1.ConfigMapLister
-	ImageConfigSynced  cache.InformerSynced
-	BuildConfigSynced  cache.InformerSynced
-	ConfigMapSynced    cache.InformerSynced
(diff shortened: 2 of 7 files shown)


deads2k:
/hold

@bparees sorry, read the wrong repo.

pkg/operator/configobservation/interfaces.go
```diff
@@ -12,8 +12,6 @@ type Listers struct {
 	ImageConfigLister  configlistersv1.ImageLister
 	BuildConfigLister  configlistersv1.BuildLister
 	ConfigMapLister    corelistersv1.ConfigMapLister
-	ImageConfigSynced  cache.InformerSynced
-	BuildConfigSynced  cache.InformerSynced
 	ConfigMapSynced    cache.InformerSynced
```
bparees: why do we still have this one?

bparees:
this just looks like refactoring, what does it have to do w/ cluster-config-operator?

sttts:
+1 to what @bparees wrote.

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -31,11 +31,6 @@ func ObserveInternalRegistryHostname(genericListers configobserver.Listers, reco
 		}
 	}
 
-	if !listers.ImageConfigSynced() {
-		glog.Warning("images.config.openshift.io not synced")
-		return prevObservedConfig, errs
-	}
-
```
mfojtik: couple lines below we have:

```
	if errors.IsNotFound(err) {
		glog.V(2).Infof("images.config.openshift.io/cluster: not found")
		return observedConfig, errs
	}
```

shouldn't this be `prevObservedConfig` ?

bparees: no...  if you don't have an images cluster config object, then we should return an empty observed config because you have an empty images config.

if we get a different error (ie api error) then we return the prev observed config.

bparees:
@sanchezl the changes look ok, but i'm still puzzled by the PR+commit description... please explain or fix that before this merges.

pkg/operator/configobservation/interfaces.go
bparees: (and address this question i guess)

sanchezl: Thanks @bparees . It's been removed.

bparees:
/lgtm

thanks @sanchezl

bparees:
/hold cancel

@sanchezl this is failing verification, please take a look.

bparees:
@sanchezl i fixed this up and an merging it here: https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/73

PR #72, opened 2019-02-12
rename controller operator resource instance from 'instance' to 'cluster'

bparees:
makes us consistent w/ the other controller+api operators.

### DIFF
--- a/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
+++ b/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
@@ -101,7 +101,7 @@ func syncOpenShiftControllerManager_v311_00_to_latest(c OpenShiftControllerManag
 		progressingMessages = append(progressingMessages, fmt.Sprintf("daemonset/controller-manager: observed generation is %d, desired generation is %d.", actualDaemonSet.Status.ObservedGeneration, actualDaemonSet.ObjectMeta.Generation))
 	}
 	if operatorConfig.ObjectMeta.Generation != operatorConfig.Status.ObservedGeneration {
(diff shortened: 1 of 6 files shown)


bparees:
@deads2k you're welcome

/assign @adambkaplan

bparees:
> nit - we should also update the operator progressing message here: https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go#L104

Good catch, updated.

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -101,7 +101,7 @@ func syncOpenShiftControllerManager_v311_00_to_latest(c OpenShiftControllerManag
 		progressingMessages = append(progressingMessages, fmt.Sprintf("daemonset/controller-manager: observed generation is %d, desired generation is %d.", actualDaemonSet.Status.ObservedGeneration, actualDaemonSet.ObjectMeta.Generation))
 	}
 	if operatorConfig.ObjectMeta.Generation != operatorConfig.Status.ObservedGeneration {
-		progressingMessages = append(progressingMessages, fmt.Sprintf("openshiftcontrollermanageroperatorconfigs/instance: observed generation is %d, desired generation is %d.", operatorConfig.Status.ObservedGeneration, operatorConfig.ObjectMeta.Generation))
+		progressingMessages = append(progressingMessages, fmt.Sprintf("openshiftcontrollermanageroperatorconfigs/cluster: observed generation is %d, desired generation is %d.", operatorConfig.Status.ObservedGeneration, operatorConfig.ObjectMeta.Generation))
```
adambkaplan: @bparees I think the CRD for the operator config is now is now `openshiftcontrollermanagers.operator.openshift.io` - should update this message so an admin can copy/paste to get the details.

bparees: right. updated.

adambkaplan:
/retest

Install failed - hung deploying networking

bparees:
@adambkaplan this is passing tests now

PR #75, opened 2019-02-20
bump(*)

bparees:


bparees:
/assign @adambkaplan 

picks up a fix for ensuring daemonset+deployment rollouts happen when the image field changes.

bparees:
now also picks up the fix for reading the certs on startup, from:
https://github.com/openshift/library-go/pull/267

PR #77, opened 2019-02-26
Update pod.yaml to include priorityClassName

ravisantoshgudimetla:
As of now, priorityClass is not set for openshift-controller-manager:

`oc describe pod controller-manager-7t97k -n openshift-controller-manager | grep -i priority
Priority:           0
PriorityClassName:  <none>`

I am adding system-node-critical priorityClassName so that controlle-manager is in sync in other control plane components.

### DIFF
--- a/bindata/v3.11.0/openshift-controller-manager/ds.yaml
+++ b/bindata/v3.11.0/openshift-controller-manager/ds.yaml
@@ -20,6 +20,7 @@ spec:
         app: openshift-controller-manager
         controller-manager: "true"
     spec:
+      priorityClassName: system-node-critical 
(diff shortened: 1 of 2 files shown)


coreydaley:
@adambkaplan is this something that we are interested in including?

adambkaplan:
/hold

I'm pretty sure we don't want to do this.

adambkaplan:
/hold cancel

We're going to follow oca-o and mark the openshift controller-manager as being node-critical. See https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/bindata/v3.11.0/openshift-apiserver/ds.yaml#L24

PR #78, opened 2019-02-26
bump(*): pick up https://github.com/openshift/library-go/pull/284

bparees:


bparees:
@gabemontero @adambkaplan bump

bparees:
> Interesting that only the glide.lock was picked up, and not whatever associated glide.yaml change accompanied it

there was no glide.yaml change, i was just pulling in updated versions of the deps from the repos/branches we point to.

gabemontero:
> there was no glide.yaml change, i was just pulling in updated versions of
> the deps from the repos/branches we point to.
>
I guess I would expect those deps to cut new versions as things change, but
in retrospect, sure, they could be bug fixes
to an existing tag, etc.

PR #81, opened 2019-03-04
Change operator pull spec to IfNotPresent

adambkaplan:
/assign @bparees 
/cc @smarterclayton @jwforres


adambkaplan:
Note that the operand is already using `IfNotPresent` in its pull spec: https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/bindata/v3.11.0/openshift-controller-manager/ds.yaml#L27

PR #85, opened 2019-04-01
Failure to upgrade

coreydaley:
https://bugzilla.redhat.com/show_bug.cgi

### DIFF
--- a/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
+++ b/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
@@ -80,9 +80,6 @@ func syncOpenShiftControllerManager_v311_00_to_latest(c OpenShiftControllerManag
 		errors = append(errors, fmt.Errorf("%q: %v", "deployment", err))
 	}
 
-	operatorConfig.Status.ObservedGeneration = operatorConfig.ObjectMeta.Generation
-	resourcemerge.SetDaemonSetGeneration(&operatorConfig.Status.Generations, actualDaemonSet)
-
 	// manage status
 	if actualDaemonSet.Status.NumberAvailable > 0 {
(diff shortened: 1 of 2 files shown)


adambkaplan:
@coreydaley I tend to prefer that all vendor code updates should happen in the bump(*) commit, and any change thereafter has its own commit(s) - even if that means the bump results in broken/non-compiling code. It helps keep the history clean.

coreydaley:
@adambkaplan fixed, must have over-squashed before my last push.

adambkaplan:
/hold

per @mfojtik

glide.lock
```diff
@@ -474,7 +474,7 @@ imports:
   - storage/v1alpha1
   - storage/v1beta1
 - name: k8s.io/apiextensions-apiserver
-  version: 63bd5e1c39483f0ef7fd275794d19a7cba19b764
+  version: 36be1826c980fa7ca82d91645e06e3157f159322
```
adambkaplan: @coreydaley We should pin this (and any other kube deps) to a 1.12 branch. 

Let's split this out into a separate PR and update this PR to only contain the changes around progress reporting in the sync controller.

coreydaley: @adambkaplan Aren't they using 1.12 in the glide.yaml file? https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/glide.yaml ?

coreydaley:
/retest

level=error msg="\t* aws_vpc.new_vpc: Error creating VPC: VpcLimitExceeded: The maximum number of VPCs has been reached."

adambkaplan:
@coreydaley the openshift-apiserver-operator has unit test patterns we can take advantage of for this change: https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/pkg/operator/workloadcontroller/workload_controller_openshiftapiserver_v311_00_test.go#L33-L173

adambkaplan:
/hold cancel

We aren't bumping deps in this PR any more

PR #92, opened 2019-04-22
Use the CVO for operator config management

damemi:
Don't create operator resource with EnsureOperatorConfigExists, which is being removed, and instead tag the manifest to be managed by the CVO


damemi:
bump @adambkaplan can you please approve this?

PR #94, opened 2019-04-26
allow lease to be held in openshift-controller-manager

deads2k:
required for https://github.com/openshift/origin/pull/22686

The lease should move out of kube-system

/assign @adambkaplan


adambkaplan:
@deads2k how will lease be transitioned on upgrade?

deads2k:
/retest

@adambkaplan if we merge it now, no upgrade

PR #96, opened 2019-05-01
Bug 1704874: Create ConfigMap for Registry CA

adambkaplan:
Create a ConfigMap that the openshift-controller-manager can use to hold the registry's CA.
Use the service.beta.openshift.io/inject-cabundle annotation to have the service-ca-operator dynamically inject the internal registry's certificate authority.

See https://bugzilla.redhat.com/show_bug.cgi?id=1704874

### DIFF
--- a/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
+++ b/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
@@ -8,6 +8,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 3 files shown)


adambkaplan:
@bparees I think we'll want the operator to create the ConfigMap to hold the CA for the registry.
Placing it in the controller-manager's namespace seems more natural (at least I'm certain we won't run into RBAC issues accessing the ConfigMap)

bparees:
> @bparees I think we'll want the operator to create the ConfigMap to hold the CA for the registry.
Placing it in the controller-manager's namespace seems more natural

since it's something consumed by the controllers, that seems reasonable to me.

adambkaplan:
ping @bparees 

/cc @gabemontero @coreydaley

bindata/v3.11.0/openshift-controller-manager/openshift-registry-ca-cm.yaml
```diff
@@ -0,0 +1,8 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: openshift-registry-ca
```
bparees: this is the openshift-service-ca, it's not registry specific.

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -189,6 +195,38 @@ func manageOpenShiftControllerManagerConfigMap_v311_00_to_latest(kubeClient kube
...
+		}
+		recorder.Eventf("ConfigMapCreated", "Created %s%s/%s%s because it was missing", "configmap", "", "openshift-registry-ca", "-n openshift-controller-manager")
+		return true, nil
+	}
+
+	// Ensure the openshift-registry-ca ConfigMap has the service.beta.openshift.io/inject-cabundle annotation
+	// Otherwise ignore the contents of the ConfigMap
+	modified := resourcemerge.BoolPtr(false)
+	existingCopy := existing.DeepCopy()
+	resourcemerge.EnsureObjectMeta(modified, &existingCopy.ObjectMeta, configMap.ObjectMeta)
```
bparees: how are you checking whether the service-ca value within the configmap has changed?

in our other configmaps we store a hash of the keys/values so we know if something has changed that needs to force a rollout of the configmap consumer....

bparees: you should be able to just add that in manageOpenShiftControllerManagerConfigMap.  In fact isn't that already watching a client-ca configmap?  (maybe for a differnet ca)

adambkaplan: yes, we are watching the client-ca ConfigMap and the serving-cert secret.
We'll still need this function to create the `openshift-service-ca` ConfigMap and ensure that it has the `inject-cabundle` annotation.

bparees: oh. you do not need to watch the openshift-service-ca configmap from here because the controller-manager itself does not care if that configmap changes.  only the buildcontroller logic inside the controllermanager, which already watches the configmap itself.

adambkaplan: @bparees renamed ConfigMap to `openshift-service-ca` - ptal

PR #106, opened 2019-07-10
Migrate to build machinery

tnozicka:
- other operators use build machinery from library-go
- you get ldflags for version, verify-deps and other benefits
- we keep the shared code updated in one repo (library-go)

/assign @adambkaplan 

replaces: https://github.com/openshift/release/pull/4308


sttts:
/hold

Leaving finally decission to @adambkaplan

tnozicka:
/hold
for @adambkaplan to bless it

manifests/02_config.crd.yaml-merge-patch
```diff
@@ -0,0 +1,10 @@
+# this file can be removed once we switch to v0.2 of crd generator
+# see: https://github.com/openshift/cluster-kube-scheduler-operator/pull/148
```
sttts: @damemi stange comment. I don't see the pattern annotation in the api types. Maybe we should add it and get rid of these patches.

damemi: https://github.com/openshift/api/blob/master/operator/v1/types.go#L27 ?

PR #105, opened 2019-07-08
Use new crd-schema-gen tool

damemi:
This PR:
1. Updates Makefile to use new, updated, containerized crd-schema-gen
2. Removes old deps based on forks of upstream deps for the old crd-gen scripts
3. Bumps deps for API changes and to clear out deps that were just removed
4. Updates CRD to pass verify

PR #107, opened 2019-07-11
Add spec to required fields in manifest

damemi:
https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/105 erroneously removed spec from the required fields. This bumps and re-adds it


sttts:
/hold

@adambkaplan for final lgtm.

PR #108, opened 2019-07-17
use the new "openshift-controller-manager" image

sanchezl:
Depends on the `openshift-controller-manager` being published.


sttts:
Failing tests looked unrelated.

tnozicka:
I don't think ocm would be responsible for template flakes

smarterclayton:
This was force merged even with failing tests?  It looks like this has the same failures we've been seeing all day?

PR #116, opened 2019-08-23
Bug 1719188: set library-go bits to declare we do not support unmanaged/removed; allow all other reconciliaition as part of ignoring unmanaged/removed; clean up messages around not supporting unmanaged/removed

gabemontero:


pkg/operator/operator.go
```diff
@@ -78,6 +82,7 @@ func (c OpenShiftControllerManagerOperator) sync() error {
 		return err
 	}
 	switch operatorConfig.Spec.ManagementState {
+	// if we decide to not allow this operator to be set unmanaged, then remove the management.SetOperatorAlwaysManaged() call in NewOpenShiftControllerManagerOperator()
 	case operatorapiv1.Unmanaged:
```
gabemontero: Is there a chance we do *NOT* want this operator to be unmanaged/removed @bparees @adambkaplan ?

pkg/operator/operator.go
```diff
@@ -69,6 +70,9 @@ func NewOpenShiftControllerManagerOperator(
 	// we only watch some namespaces
 	kubeInformersForOpenshiftControllerManager.Core().V1().Namespaces().Informer().AddEventHandler(c.namespaceEventHandler())
 
+	// set this bit so the library-go code employed for clusteroperator reconciliation allows this operator to be set to unmanaged
```
bparees:
```suggestion
	// set this bit so the library-go code employed for clusteroperator reconciliation does not allow this operator to be set to unmanaged
```

pkg/operator/operator.go
```diff
@@ -69,6 +70,9 @@ func NewOpenShiftControllerManagerOperator(
...
+	management.SetOperatorAlwaysManaged()
```
bparees: wait..which direction are you trying to go?

bparees: Calling this means you will *not* be able to set the operator to "unmanaged".

```
// SetOperatorAlwaysManaged is one time choice when an operator want to opt-out from supporting the "unmanaged" state.
// This is a case of control plane operators or operators that are required to always run otherwise the cluster will
// get into unstable state or critical components will stop working.
func SetOperatorAlwaysManaged() {
	allowOperatorUnmanagedState = false
}
```

pkg/operator/operator.go
bparees: we do not want it to be removed for sure. Doing so will apparently break lots of stuff because things like the SCC controller run inside this pod and if it's not there, pods can't be admitted.

/cc @deads2k 

Presumably we also don't want to allow unmanaged, for similar reasons, but that's less clear to me since it might be needed/useful for manual intervention on a cluster.

pkg/operator/operator.go
gabemontero: Yeah perhaps see my comment https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/116#discussion_r317266003 perhaps

we need this call if we want to allow the operator to process getting set to unmanaged/removed, otherwise the library-go code will set the cluster operator conditions to UNKNOWN

if we decide the OCM operator should always stay in managed, then we remove this call and change the case statement starting at https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/116/files#diff-554de5523753fda8c93e7008c9bd947fR84 to reject / ignore such a spec change.

pkg/operator/operator.go
gabemontero: no ... it is allows, not "does not allow"

pkg/operator/operator.go
adambkaplan: Based on the godoc comment I say we want to do this for ocm-o. Without it most of what makes OpenShift "OpenShift" breaks.

pkg/operator/operator.go
adambkaplan: @gabemontero my vote is that this operator should _never_ be removed. It's also hard for me to envision a situation where a customer would need to set this to Unmanaged, but keep other control plane components managed.
@bparees or @derekwaynecarr may have a different take.

pkg/operator/operator.go
gabemontero: OK @bparees and I had a side confab.  The net is:
a) we are going to employ `SetOperatorAlwaysManaged()`
b) we are not going to accept unmanaged/removed on the ocm object any longer

gabemontero:
OK I've pivoted on this PR @adambkaplan @bparees based on discussions here as well as in slack.  To summarize the changes:
- we no longer support setting both Unmanaged/Removed on the openshiftcontrollermanager object, as such, that means
-- we declare to library-go as such, so that its sync loops react accordingly .. as such, it will not override status and set Unknown
-- not supporting means totally ignoring, so we still go through the OCM sync to process any changes
-- we add reason/messages across all the conditions (yes I know probably overkill, but we can decide which conditions to set the message on during review) indicating we do not support unmanaged/removed and that the setting has no impact on status/processing

PTAL

pkg/operator/operator.go
```diff
@@ -77,43 +82,34 @@ func (c OpenShiftControllerManagerOperator) sync() error {
...
+		// as we are ignoring / not supporting unmanaged/removed, we still process any other inputs to the sync/reconciliation
+		// unlike what we used to do
+	case operatorapiv1.Managed:
+		// we want to empty out the reason/message string if transitioning from the other phases so the default setting above is good
+	}
+
+	for _, condition := range operatorConfig.Status.Conditions {
+		// do not change the current status as part of noting that unmanaged/removed are not supported
+		condition.Reason = reasonString
+		condition.Message = messageString
```
bparees: This should not clear the previous message/reason.

if i set the management state to "unmanaged", and it's being ignored, i should not lose the pre-existing reasons/messages for whatever condition my operator is in (or enters in the future).

gabemontero: that is not what will happen here, nor what I saw in my testing

Though we are "ignoring" unmanaged/removed, it stays set in the Spec.  So we will reset the reason/message with the values we want for unmanaged/removed each time ... on each event.

The code below simply will not issue another updatestatus if the fields did not change since the last pass.

If that doesn't address your concern, I'm either not understanding your concern or missing how your concern enters the mix here.

bparees: My reading of the code is that the behavior will be this:

starting state:  managed
available=true, reason=Normal, message=everything is great

i change it to state: unmanaged, the condition reasons/message will change to:
available=true, reason="unmanaged is being ignored", message="unmanaged is being ignored"

is that right?  If so, that is what i think is incorrect behavior.  Yes i need to know my attempt to set the management state was ignored, but i should also continue to be told the actual reasons/message for the actual state the operator is in.

gabemontero: Reason/Message are blank when managed and available=true and everything is good.

And in fact, when examining `syncOpenShiftControllerManager_v311_00_to_latest`, the only time Reason/Message are set is when there is an error.

And `syncOpenShiftControllerManager_v311_00_to_latest` will override the "unmanaged is being ignored" message/reason we set here if in fact it finds something wrong 

So this really is just a replacement of blank reason/message, not a replacement of non-blank reason/message

bparees: ok

PR #118, opened 2019-09-03
create configmap for global ca injection for use by builds

gabemontero:
/assign @bparees 

@openshift/openshift-team-developer-experience FYI

per recent discussion on devex and warroom-proxy slack around https://bugzilla.redhat.com/show_bug.cgi?id=1748271

This would provide for an optimization in build controller

build controller in openshift-controller-manager could then just read in this data like it does for the service serving CA vs. labelling the config map it creates for mounting in the build pod

### DIFF
--- a/bindata/v3.11.0/openshift-controller-manager/openshift-global-ca-cm.yaml
+++ b/bindata/v3.11.0/openshift-controller-manager/openshift-global-ca-cm.yaml
@@ -1,0 +1,8 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: openshift-global-ca
+  namespace: openshift-controller-manager
+  labels: 
+    config.openshift.io/inject-trusted-cabundle: "true"
+data: {}
--- a/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
+++ b/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
@@ -78,8 +78,13 @@ func syncOpenShiftControllerManager_v311_00_to_latest(c OpenShiftControllerManag
 		errors = append(errors, fmt.Errorf("%q: %v", "openshift-service-ca", err))
 	}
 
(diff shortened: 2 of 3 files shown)


manifests/03_configmap.yaml
```diff
@@ -7,3 +7,14 @@ data:
...
+---
+
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: openshift-global-ca
+  namespace: openshift-controller-manager
+  labels: 
+    config.openshift.io/inject-trusted-cabundle: "true"
+data: {}
```
bparees: this file is not a resource list, so you can't just add another resource like this.

the service-ca CM is managed as bindata, seen here:
https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/bindata/v3.11.0/openshift-controller-manager/openshift-service-ca-cm.yaml

https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go#L76

https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go#L198-L228

so that's the pattern to copy.

gabemontero: yeah I waffled on which way to go

I convinced myself that the service-ca stuff existed in 3.11, and that is why it made sense for it to go there

bparees: That code isn't really related to 3.11... all those "311" named things are leftovers from when it was assumed we'd be migrating clusters from 3.11 to 4.x.  But we never used service CA injection in 3.11, we just used the CA that was auto mounted into the pods.

gabemontero:
refactor pushed @bparees

adambkaplan:
/retitle Bug 1748271: Create configmap for global CA injection within builds

gabemontero:
On Wed, Sep 4, 2019 at 11:18 AM Adam Kaplan <notifications@github.com>
wrote:

> /retitle Bug 1748271: Create configmap for global CA injection within
> builds
>
@adambkaplan we should retitle the associated OCM PR instead of this one

PR #130, opened 2019-11-11
monitoring: move service monitor RBAC rules to manifests/

sttts:
These are deployed by monitoring operator, but shouldn't: https://github.com/openshift/cluster-monitoring-operator/blob/e515ed03618466b6a1b9d2fcf12192191bfd90f8/assets/prometheus-k8s/role-specific-namespaces.yaml#L51-L66

### DIFF
--- a/manifests/0000_90_openshift-controller-manager-operator_03_operand-servicemonitor.yaml
+++ b/manifests/0000_90_openshift-controller-manager-operator_03_operand-servicemonitor.yaml
@@ -1,3 +1,34 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: prometheus-k8s
+  namespace: openshift-controller-manager
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - services
+  - endpoints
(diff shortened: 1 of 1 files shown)


sttts:
@s-urbaniak please tag

manifests/0000_90_openshift-controller-manager-operator_03_operand-servicemonitor.yaml
```diff
@@ -1,3 +1,34 @@
...
+  name: prometheus-k8s
+  namespace: openshift-controller-manager
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: prometheus-k8s
+subjects:
+- kind: ServiceAccount
+  name: prometheus-k8s
+  namespace: openshift-monitor
```
s-urbaniak: oh wait, since I fixed this in other places, this should be `openshift-monitoring`

sttts: fixed

PR #138, opened 2020-02-04
Move default config to its own path

marun:
/cc @sttts


adambkaplan:
@marun this can be squashed into a single commit.

Note that since we are past feature freeze, we need a Bugzilla to merge any PR.

marun:
@adambkaplan If you care to have this fix in this release, maybe you could file the BZ? Otherwise it could wait. I'm just trying to be helpful here, not make work for myself or anyone else.

marun:
/hold

Waiting for 4.5, this is not worth filing a bz for.

PR #139, opened 2020-02-19
Bug 1804432: Fallback to logs on ocm termination

adambkaplan:
Update ocm's DaemonSet to fallback to logs on failed termination.


adambkaplan:
/assign @deads2k 

fyi @openshift/openshift-team-developer-experience

PR #145, opened 2020-03-11
Bug 1805168: inject global proxy envs,ca into OCM daemonset

gabemontero:
per recent discussion between @dmage, @adambkaplan, and myself
the image signature controller in OCM is reaching out to the registries like registry.redhat.io and not using the global proxy if set

see https://github.com/openshift/cluster-version-operator/blob/611490ff448962c70ab29a90764a6c30d4ee87dc/lib/resourcebuilder/apps.go#L55

the bug originator manually set the proxy ENVs on the OCM and saw things works

@bparees @mfojtik @soltysh FYI

### DIFF
--- a/bindata/v3.11.0/openshift-controller-manager/ds.yaml
+++ b/bindata/v3.11.0/openshift-controller-manager/ds.yaml
@@ -45,6 +45,8 @@ spec:
           name: client-ca
         - mountPath: /var/run/secrets/serving-cert
           name: serving-cert
+        - mountPath: /etc/pki/ca-trust/extracted/pem
+          name: proxy-ca-bundles
       volumes:
       - name: config
         configMap:
@@ -55,8 +57,13 @@ spec:
       - name: serving-cert
         secret:
           secretName: serving-cert
+      - name: proxy-ca-bundles
+        configMap:
+          name: openshift-global-ca
+          items:
+            - key: ca-bundle.crt
+              path: tls-ca-bundle.pem
       nodeSelector:
         node-role.kubernetes.io/master: ""
-      priorityClassName: "system-cluster-critical"
       tolerations:
       - operator: Exists
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -18,6 +18,8 @@ import (
 	"k8s.io/klog"
 
 	operatorapiv1 "github.com/openshift/api/operator/v1"
+	configinformerv1 "github.com/openshift/client-go/config/informers/externalversions/config/v1"
+	proxyvclient1 "github.com/openshift/client-go/config/listers/config/v1"
 	operatorclientv1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
 	operatorinformersv1 "github.com/openshift/client-go/operator/informers/externalversions/operator/v1"
 	"github.com/openshift/cluster-openshift-controller-manager-operator/pkg/util"
@@ -34,6 +36,7 @@ const (
 type OpenShiftControllerManagerOperator struct {
 	targetImagePullSpec  string
 	operatorConfigClient operatorclientv1.OperatorV1Interface
+	proxyLister          proxyvclient1.ProxyLister
 
 	kubeClient kubernetes.Interface
 
@@ -47,6 +50,7 @@ type OpenShiftControllerManagerOperator struct {
 func NewOpenShiftControllerManagerOperator(
 	targetImagePullSpec string,
 	operatorConfigInformer operatorinformersv1.OpenShiftControllerManagerInformer,
+	proxyInformer configinformerv1.ProxyInformer,
 	kubeInformersForOpenshiftControllerManager informers.SharedInformerFactory,
 	operatorConfigClient operatorclientv1.OperatorV1Interface,
 	kubeClient kubernetes.Interface,
@@ -55,6 +59,7 @@ func NewOpenShiftControllerManagerOperator(
 	c := &OpenShiftControllerManagerOperator{
 		targetImagePullSpec:  targetImagePullSpec,
 		operatorConfigClient: operatorConfigClient,
+		proxyLister:          proxyInformer.Lister(),
 		kubeClient:           kubeClient,
 		queue:                workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "KubeApiserverOperator"),
 		rateLimiter:          flowcontrol.NewTokenBucketRateLimiter(0.05 /*3 per minute*/, 4),
@@ -62,6 +67,7 @@ func NewOpenShiftControllerManagerOperator(
 	}
 
 	operatorConfigInformer.Informer().AddEventHandler(c.eventHandler())
+	proxyInformer.Informer().AddEventHandler(c.eventHandler())
 	kubeInformersForOpenshiftControllerManager.Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
 	kubeInformersForOpenshiftControllerManager.Core().V1().ServiceAccounts().Informer().AddEventHandler(c.eventHandler())
 	kubeInformersForOpenshiftControllerManager.Core().V1().Services().Informer().AddEventHandler(c.eventHandler())
--- a/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
+++ b/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
@@ -3,6 +3,7 @@ package operator
 import (
 	"fmt"
 	"os"
+	"reflect"
 	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -16,6 +17,9 @@ import (
 	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
 
 	operatorapiv1 "github.com/openshift/api/operator/v1"
+
+	proxyvclient1 "github.com/openshift/client-go/config/listers/config/v1"
+
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 	"github.com/openshift/library-go/pkg/operator/resource/resourcehash"
@@ -89,7 +93,7 @@ func syncOpenShiftControllerManager_v311_00_to_latest(c OpenShiftControllerManag
 
 	// our configmaps and secrets are in order, now it is time to create the DS
 	// TODO check basic preconditions here
-	actualDaemonSet, _, err := manageOpenShiftControllerManagerDeployment_v311_00_to_latest(c.kubeClient.AppsV1(), c.recorder, operatorConfig, c.targetImagePullSpec, operatorConfig.Status.Generations, forceRollout)
+	actualDaemonSet, _, err := manageOpenShiftControllerManagerDeployment_v311_00_to_latest(c.kubeClient.AppsV1(), c.recorder, operatorConfig, c.targetImagePullSpec, operatorConfig.Status.Generations, forceRollout, c.proxyLister)
 	if err != nil {
 		errors = append(errors, fmt.Errorf("%q: %v", "deployment", err))
 	}
@@ -196,6 +200,7 @@ func manageOpenShiftControllerManagerConfigMap_v311_00_to_latest(kubeClient kube
 		kubeClient,
 		resourcehash.NewObjectRef().ForConfigMap().InNamespace(util.TargetNamespace).Named("client-ca"),
 		resourcehash.NewObjectRef().ForSecret().InNamespace(util.TargetNamespace).Named("serving-cert"),
+		resourcehash.NewObjectRef().ForConfigMap().InNamespace(util.TargetNamespace).Named("openshift-global-ca"),
 	)
 	if err != nil {
 		return nil, false, err
@@ -271,7 +276,7 @@ func manageOpenShiftGlobalCAConfigMap_v311_00_to_latest(kubeClient kubernetes.In
 	return updated, true, nil
 }
 
-func manageOpenShiftControllerManagerDeployment_v311_00_to_latest(client appsclientv1.DaemonSetsGetter, recorder events.Recorder, options *operatorapiv1.OpenShiftControllerManager, imagePullSpec string, generationStatus []operatorapiv1.GenerationStatus, forceRollout bool) (*appsv1.DaemonSet, bool, error) {
+func manageOpenShiftControllerManagerDeployment_v311_00_to_latest(client appsclientv1.DaemonSetsGetter, recorder events.Recorder, options *operatorapiv1.OpenShiftControllerManager, imagePullSpec string, generationStatus []operatorapiv1.GenerationStatus, forceRollout bool, proxyLister proxyvclient1.ProxyLister) (*appsv1.DaemonSet, bool, error) {
 	required := resourceread.ReadDaemonSetV1OrDie(v311_00_assets.MustAsset("v3.11.0/openshift-controller-manager/ds.yaml"))
 
 	if len(imagePullSpec) > 0 {
@@ -294,6 +299,61 @@ func manageOpenShiftControllerManagerDeployment_v311_00_to_latest(client appscli
 		required.Annotations = map[string]string{}
 	}
 	required.Annotations[util.VersionAnnotation] = os.Getenv("RELEASE_VERSION")
+
+	proxyCfg, err := proxyLister.Get("cluster")
+	if err != nil {
+		recorder.Eventf("ProxyConfigGetFailed", "Error retrieving global proxy config: %s", err.Error())
+		if !apierrors.IsNotFound(err) {
+			// return daemonset since it is still referenced by caller even with errors
+			return required, false, err
+		}
+	} else {
+		for i, c := range required.Spec.Template.Spec.Containers {
+			newEnvs := []corev1.EnvVar{}
+
+			if len(c.Env) == 0 {
+				if len(proxyCfg.Status.NoProxy) > 0 {
+					newEnvs = append(newEnvs, corev1.EnvVar{Name: "NO_PROXY", Value: proxyCfg.Status.NoProxy})
+				}
+				if len(proxyCfg.Status.HTTPProxy) > 0 {
+					newEnvs = append(newEnvs, corev1.EnvVar{Name: "HTTP_PROXY", Value: proxyCfg.Status.HTTPProxy})
+				}
+				if len(proxyCfg.Status.HTTPSProxy) > 0 {
+					newEnvs = append(newEnvs, corev1.EnvVar{Name: "HTTPS_PROXY", Value: proxyCfg.Status.HTTPSProxy})
+				}
+			}
+
+			for _, env := range c.Env {
+				name := strings.TrimSpace(env.Name)
+				switch name {
+				case "HTTPS_PROXY":
+					if len(proxyCfg.Status.HTTPSProxy) == 0 {
+						continue
+					}
+					env.Value = proxyCfg.Status.HTTPSProxy
+
+				case "HTTP_PROXY":
+					if len(proxyCfg.Status.HTTPProxy) == 0 {
+						continue
+					}
+					env.Value = proxyCfg.Status.HTTPProxy
+
+				case "NO_PROXY":
+					if len(proxyCfg.Status.NoProxy) == 0 {
+						continue
+					}
+					env.Value = proxyCfg.Status.NoProxy
+
+				}
+				newEnvs = append(newEnvs, env)
+			}
+			// reflect.DeepEqual does not consider this case equal
+			envsEqual := c.Env == nil && len(newEnvs) == 0
+			envsEqual = envsEqual || !reflect.DeepEqual(newEnvs, c.Env)
+			forceRollout = forceRollout || !envsEqual
+			required.Spec.Template.Spec.Containers[i].Env = newEnvs
+		}
+	}
 
 	return resourceapply.ApplyDaemonSet(client, recorder, required, resourcemerge.ExpectedDaemonSetGeneration(required, generationStatus), forceRollout)
 }
--- a/pkg/operator/sync_openshiftcontrollermanager_v311_00_test.go
+++ b/pkg/operator/sync_openshiftcontrollermanager_v311_00_test.go
@@ -1,6 +1,9 @@
 package operator
 
 import (
+	configv1 "github.com/openshift/api/config/v1"
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
+	"k8s.io/client-go/tools/cache"
 	"testing"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -157,6 +160,9 @@ func TestProgressingCondition(t *testing.T) {
 					},
 				})
 
+			indexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
+			proxyLister := configlistersv1.NewProxyLister(indexer)
+
 			operatorConfig := &operatorv1.OpenShiftControllerManager{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:       "cluster",
@@ -175,6 +181,7 @@ func TestProgressingCondition(t *testing.T) {
 
 			operator := OpenShiftControllerManagerOperator{
 				kubeClient:           kubeClient,
+				proxyLister:          proxyLister,
 				recorder:             events.NewInMemoryRecorder(""),
 				operatorConfigClient: controllerManagerOperatorClient.OperatorV1(),
 			}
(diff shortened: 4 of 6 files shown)


gabemontero:
existing e2e's are green at least

/assign @bparees 
/assign @adambkaplan

bparees:
I don't think this is the right solution.

the CVO injection is only for operators.  As in "I need the CVO to inject these env vars into my operators' deployment because the CVO owns my deployment and only it can touch it".

You(the OCM operator) own the OCM's deployment, you should be setting the proxy envs on the OCM's deployment and reconciling them as they change (i.e. OCM operator needs to watch the proxy config and propagate that config to the OCM deployment)

gabemontero:
> I don't think this is the right solution.
> 
> the CVO injection is only for operators. As in "I need the CVO to inject these env vars into my operators' deployment because the CVO owns my deployment and only it can touch it".
> 
> You(the OCM operator) own the OCM's deployment, you should be setting the proxy envs on the OCM's deployment and reconciling them as they change (i.e. OCM operator needs to watch the proxy config and propagate that config to the OCM deployment)

OK a bigger change but certainly not onerours.  @adambkaplan 's suggestion of the annotation was trivial enough that at least proposing it to a bigger audience for evaluation seemed fine.

There is also the possibility of the OCM, which is already watching proxy config, setting the envs on itself.  While less expensive, that probably subverts our "what an operator does and what an operand does" conventions.  But certainly voice any opinions there @bparees if you like.

bparees:
> There is also the possibility of the OCM, which is already watching proxy config, setting the envs on itself. While less expensive, that probably subverts our "what an operator does and what an operand does" conventions. But certainly voice any opinions there @bparees if you like.

the resource needs to be owned/written to by a single writer.  for the OCM daemonset resource, that's the OCM operator.  Otherwise you risk two things fighting w/ each other about what the correct state of the resource should be.

gabemontero:
> > There is also the possibility of the OCM, which is already watching proxy config, setting the envs on itself. While less expensive, that probably subverts our "what an operator does and what an operand does" conventions. But certainly voice any opinions there @bparees if you like.
> 
> the resource needs to be owned/written to by a single writer. for the OCM daemonset resource, that's the OCM operator. Otherwise you risk two things fighting w/ each other about what the correct state of the resource should be.

Well I was actually thinking it would be golang `os.Setenv()` call in the OCM process actually nad would not be manipulating the DaemonSet at all.  It would be dynamically updating the env's the DaemonSet added to the underlying Pod.

But again, probably too off the beaten path ... I just mentioned it cause it popped into my head

I'll start down the path in this repo

bparees:
> Well I was actually thinking it would be golang os.Setenv() call in the OCM process actually nad would not be manipulating the DaemonSet at all. It would be dynamically updating the env's the DaemonSet added to the underlying Pod.

the drawbacks w/ that approach are:
1) lack of debuggability (you can't just look at the resource to understand what configuration it is using)
2) I believe the transports+static libs get setup early and read this information just once, so changing the env var will not, in many cases, affect the configuration they are using, so you'd need to restart the process to pick up the change anyway, which means the OCM would have to nuke itself, and then ensure(on restart) the envs got set before any lib+transport initialization

(2) is why we had to make the CVO able to inject proxy config on operator resources, because operators could not solve the problem themselves by dynamically setting their own env vars.  You're basically in the same boat, one level deeper.

bparees:
Note: i think i may have seen another devex PR recently that was adding the CVO proxy injection annotation.  I didn't look at it closely... but if it was doing the same thing as this one (adding the injection annotation to an operand resource, not an operator resource) it should be reverted and redone per the approach i've outlined in the discussion here.

gabemontero:
> Note: i think i may have seen another devex PR recently that was adding the CVO proxy injection annotation. I didn't look at it closely... but if it was doing the same thing as this one (adding the injection annotation to an operand resource, not an operator resource) it should be reverted and redone per the approach i've outlined in the discussion here.

since @adambkaplan suggested the annotation to me, maybe he know which PR you are referring to @bparees 

otherwise, I'll refactor this PR to have OCM-O watch global proxy and update the OCM daemonset's env as you previously mentioned in the classic operator/operand mgmt pattern

bparees:
btw if the operand has to consume the proxy it also needs to consume the proxy CAs, so there's more that needs to be done here.

the operand has to mount a configmap that's annotated for the proxy CA injection, and the operator has to watch that configmap and redeploy the operand if the configmap is updated w/ new CA content.

gabemontero:
> btw if the operand has to consume the proxy it also needs to consume the proxy CAs, so there's more that needs to be done here.
> 
> the operand has to mount a configmap that's annotated for the proxy CA injection, and the operator has to watch that configmap and redeploy the operand if the configmap is updated w/ new CA content.

yep true as well ... so yeah the OCM-O is already creating a proxy injected CM for the CAs and is watching it as part of the build support ... the daemonset update would have to include mounting that CA in addition to the envs

And I suppose the OCM would have to be updated to copy the cert from the k8s mount point to the well known location under /etc just like the openshift/builder image does today

bparees:
> And I suppose the OCM would have to be updated to copy the cert from the k8s mount point to the well known location under /etc just like the openshift/builder image does today

yes, or mount the CM there directly.  (The CM should contain all the CAs you need, including system trusts, so it should be ok to just mount over top of the image's own CAs)

adambkaplan:
> yes, or mount the CM there directly. (The CM should contain all the CAs you need, including system trusts, so it should be ok to just mount over top of the image's own CAs)

I concur that mounting directly is the way to go for OCM. Builds are kind of a special case because we have `additionalTrustedCAs` that are used only for pulling/pushing images. Same deal with the image registry.

soltysh:
I think you need something similar to https://github.com/openshift/cluster-kube-controller-manager-operator/pull/285 and  https://github.com/openshift/cluster-kube-controller-manager-operator/pull/325 to be able to consume and use the proxy as far as I can tell.

gabemontero:
> I think you need something similar to [openshift/cluster-kube-controller-manager-operator#285](https://github.com/openshift/cluster-kube-controller-manager-operator/pull/285) and [openshift/cluster-kube-controller-manager-operator#325](https://github.com/openshift/cluster-kube-controller-manager-operator/pull/325) to be able to consume and use the proxy as far as I can tell.

thanks for the pointers @soltysh 

I've got some changes on my laptop that share *some* commonality with those PRs

I'm wrapping up some unit tests and should have a PR up before my EOB close today.

gabemontero:
/hold cancel

PR switched over to a proxy watch, config observation, and injection of envs/ca into OCM daemonset via operator

ptal

gabemontero:
all green e2e's at least

bump @adambkaplan @bparees

pkg/operator/operator.go
```diff
@@ -55,6 +58,7 @@ func NewOpenShiftControllerManagerOperator(
 	c := &OpenShiftControllerManagerOperator{
 		targetImagePullSpec:  targetImagePullSpec,
 		operatorConfigClient: operatorConfigClient,
+		proxyLister:          proxyLister,
```
bparees: don't you also need to add an event handler to the proxy informer below so you get notified on updates to the proxyconfig?

bparees: ah, that appears to be handled in observeconfigcontroller

bparees:
i'm missing where this:
1) watches the global CA configmap for changes (must rollout the operand when it changes)
2) watches the proxyconfig for changes to the global CA configmap name/reference (would have to change which configmap (1) is watching when it changes).

gabemontero:
> i'm missing where this:
> 
>     1. watches the global CA configmap for changes (must rollout the operand when it changes)

Yep forgot to add this (I remember thinking of it at one point) ... need an observer_configmap.go

> 
>     2. watches the proxyconfig for changes to the global CA configmap name/reference (would have to change which configmap (1) is watching when it changes).

My thought here is the OCM-O already creates a CM with the injection annotation for the OCM as part of builds support for global proxy.  As part of the new observer_configmap.go, we'd focus on watching the CM for changes.

Is that not sufficient for our needs @bparees ?

bparees:
> My thought here is the OCM-O already creates a CM with the injection annotation for the OCM as part of builds support for global proxy. As part of the new observer_configmap.go, we'd focus on watching that CM for changes.

Sorry, I think I was mixed up.  I was thinking about the configmap that the proxyconfig references (the global CM the admin edits).  But you don't have a need to watch that configmap yourself, because of what you said.

So yeah, just watching your annotated configmap for changes and forcing a rollout when that happens is sufficient.  Most teams do that by hashing the CM and putting the hash as an annotation on the deployment resource.  That way if/when the hash changes, it auotmatically triggers a rollout.

gabemontero:
> > My thought here is the OCM-O already creates a CM with the injection annotation for the OCM as part of builds support for global proxy. As part of the new observer_configmap.go, we'd focus on watching that CM for changes.
> 
> Sorry, I think I was mixed up. I was thinking about the configmap that the proxyconfig references (the global CM the admin edits). But you don't have a need to watch that configmap yourself, because of what you said.
> 
> So yeah, just watching your annotated configmap for changes and forcing a rollout when that happens is sufficient. Most teams do that by hashing the CM and putting the hash as an annotation on the deployment resource. That way if/when the hash changes, it auotmatically triggers a rollout.

Yeah I need to walk down the code paths to be sure, but I believe some of the PR refs @soltysh posted have examples of this

gabemontero:
ok pushed observe ca crt change @bparees

bparees:
this may be me forgetting how this operator works, but...while i see the observeconfig computes the CA hash and stores it in the observedconfig, i don't see where that hash is applied to the deployment to force a rollout if the hash has changed?

gabemontero:
> this may be me forgetting how this operator works, but...while i see the observeconfig computes the CA hash and stores it in the observedconfig, i don't see where that hash is applied to the deployment to force a rollout if the hash has changed?

Well I forgot the update the bindata/v3.11.0/openshift-controller-manager/ds.yaml with the CM mount while making all these other changes, so that should at least contribute to your confusion :-)

Once I do that, the flow should be:
- as with all this observed_*.go stuff, any delta in the pervious `map[string]interface` data triggers a redeployment of the OCM
- the sync_* logic applies the updated proxy env's on the deployment
- we are always mounting the the same CM onto the OCM daemonset pod, but when the recycle occurs, the pods will pick up the latest cert in the recently updated CM (whose update we just observed in the OCM-O)

bindata/v3.11.0/openshift-controller-manager/ds.yaml
```diff
@@ -55,8 +57,13 @@ spec:
...
           secretName: serving-cert
+      - name: proxy-ca-bundles
+        configMap:
+          name: openshift-global-ca
+          items:
+            - key: ca-bundle.crt
+              path: tls-ca-bundle.pem
       nodeSelector:
         node-role.kubernetes.io/master: ""
-      priorityClassName: "system-cluster-critical"
```
gabemontero: note the built in yaml validation in GoLand figured out that this was a duplicate setting with line 25 above

bindata/v3.11.0/openshift-controller-manager/ds.yaml
```diff
@@ -45,6 +45,8 @@ spec:
           name: client-ca
         - mountPath: /var/run/secrets/serving-cert
           name: serving-cert
+        - mountPath: /etc/pki/ca-trust/extracted/pem
```
gabemontero: This duplicates how we map the same config map into build pods a la https://github.com/openshift/openshift-controller-manager/blob/master/pkg/build/controller/strategy/util.go#L549-L578

bindata/v3.11.0/openshift-controller-manager/ds.yaml
```diff
@@ -55,8 +57,13 @@ spec:
       - name: serving-cert
         secret:
           secretName: serving-cert
+      - name: proxy-ca-bundles
+        configMap:
+          name: openshift-global-ca
```
gabemontero: but rather than the local namespace CM we copy in, we use our injected CM source in the OCM namesapce, `openshift-global-ca`

gabemontero:
OK @bparees *hopefully* I've finally connected all the dots ... ptal when you get the chance

gabemontero:
failures related to my ds.yaml change:

`level=error msg="Cluster operator openshift-controller-manager Degraded is True with Workload_SyncError: WorkloadDegraded: \"deployment\": DaemonSet.apps \"controller-manager\" is invalid: spec.template.spec.containers[0].volumeMounts[4].mountPath: Invalid value: \"/etc/pki/ca-trust/extracted/pem\": must be unique\nWorkloadDegraded: "`

gabemontero:
> failures related to my ds.yaml change:
> 
> `level=error msg="Cluster operator openshift-controller-manager Degraded is True with Workload_SyncError: WorkloadDegraded: \"deployment\": DaemonSet.apps \"controller-manager\" is invalid: spec.template.spec.containers[0].volumeMounts[4].mountPath: Invalid value: \"/etc/pki/ca-trust/extracted/pem\": must be unique\nWorkloadDegraded: "`

Maybe this means we have to mount in a typical `/var..` location, and then copy to `/etc/pki/ca-trust/extracted/pem` in the main.go of OCM?  .... Similar to some of the cert copies we do in openshift/builder ?

bparees:
> as with all this observed_*.go stuff, any delta in the pervious map[string]interface data triggers a redeployment of the OCM

how?  I don't see anything in https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/145/files#diff-71c619af51132696348e484ffcd8edbfR277 that is going to cause a rollout to happen when the config changes.

gabemontero:
> > as with all this observed_*.go stuff, any delta in the pervious map[string]interface data triggers a redeployment of the OCM
> 
> how? I don't see anything in https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/145/files#diff-71c619af51132696348e484ffcd8edbfR277 that is going to cause a rollout to happen when the config changes.

As part of the existing code we have for the global ca config map in the OCM namesapce, I belive this code will already trigger:  https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go#L242-L272

I initially wondered if with this code I just linked, I even needed the new observer_config.go module I added with this PR, but sided on "it couldn't hurt" ... but I would defer to whomever thinks they know better.

bparees:
> As part of the existing code we have for the global ca config map in the OCM namesapce, I belive this code will already trigger: https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go#L242-L272

that code just ensures the configmap has the injection annotation on it.

it doesn't do anything w/ respect to watching the content of the configmap and ensuring a deployment rollout happens if the configmap content changes.  it explicitly ignores the configmap content.

gabemontero:
> > As part of the existing code we have for the global ca config map in the OCM namesapce, I belive this code will already trigger: https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go#L242-L272
> 
> that code just ensures the configmap has the injection annotation on it.
> 
> it doesn't do anything w/ respect to watching the content of the configmap and ensuring a deployment rollout happens if the configmap content changes. it explicitly ignores the configmap content.

ok, I put the code for inspecting the content in observer_config.go ... I thought rollouts of the operand could be triggered from either place

1) should I duplicate that code in the sync_*.go method, storing the hash on the configmap obj ?
2) should I get rid of the observer_config.go module ?

bparees:
take a look at how cluster-openshift-api-server-operator handles this, since it has the same structure as this operator:
https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go#L358-L366

you can also see the commit that introduced proxy support in that operator:
https://github.com/openshift/cluster-openshift-apiserver-operator/commit/39b3f4f032356fc11cad969b6a9b7e263b417ba5

and some refactoring david did to use the hashes:
https://github.com/openshift/cluster-openshift-apiserver-operator/commit/469e1e7ff19a49b65b9b1cae4534b13e31d3d7a3

gabemontero:
> take a look at how cluster-openshift-api-server-operator handles this, since it has the same structure as this operator:
> https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go#L358-L366
> 
> you can also see the commit that introduced proxy support in that operator:
> [openshift/cluster-openshift-apiserver-operator@39b3f4f](https://github.com/openshift/cluster-openshift-apiserver-operator/commit/39b3f4f032356fc11cad969b6a9b7e263b417ba5)
> 
> and some refactoring david did to use the hashes:
> [openshift/cluster-openshift-apiserver-operator@469e1e7](https://github.com/openshift/cluster-openshift-apiserver-operator/commit/469e1e7ff19a49b65b9b1cae4534b13e31d3d7a3)

thanks for the refs @bparees ... I'll start monkey seeing / monkey doing shortly

Also, I see in https://github.com/openshift/cluster-openshift-apiserver-operator/tree/master/pkg/operator/configobservation no analogous config observation for proxy or the CA config maps ... I will ditch those from this PR.

gabemontero:
And knowing what to look for now, there is something similar already in this repo for the other CAs:  https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go#L187-L207

gabemontero:
changes pushed @bparees

gabemontero:
sorry wait @bparees

gabemontero:
ok now it is reviewable @bparees

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -262,16 +265,23 @@ func manageOpenShiftGlobalCAConfigMap_v311_00_to_latest(kubeClient kubernetes.In
...
 		return existing, false, nil
 	}
-	updated, err := client.ConfigMaps(util.TargetNamespace).Update(existingCopy)
+
+	// we can embed input hashes on our main configmap to drive rollouts when they change; so we are including a hash of
+	// the global proxy CA map since the image signature controller access registries outside the cluster
+	inputHashes, err := resourcehash.MultipleObjectHashStringMapForObjectReferences(
+		kubeClient,
+		resourcehash.NewObjectRef().ForConfigMap().InNamespace(util.TargetNamespace).Named("openshift-global-ca"),
+	)
```
bparees: i don't think you need to do this (nor should you, since you're adding a random key to the CA configmap which can then end up getting mounted into the pod).

just add the configmap to the list that already exists here:
https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go#L195-L199

that code is computing a hash based on all the configmaps we care about, and storing that hash in whatever "main configmap" is and applying the change to the main configmap.  If that results is an update to mainconfigmap, force rollout is returned (based on my somewhat superficial skimming of the code).

bindata/v3.11.0/openshift-controller-manager/ds.yaml
gabemontero: Cross referencing my https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/145#issuecomment-601771069 here ... the installer does not appear to like mounting directly to `/etc/pki/ca-trust/extracted/pem`

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
gabemontero: suggestion pushed @bparees thanks

bindata/v3.11.0/openshift-controller-manager/ds.yaml
bparees: that error indicates to me that something else is already adding a mount to that path to the DS.  Not understanding why you think it's related to the installer though.  That error is:

1) installer complains that the controller manager opeator is reporting degraded
2) controller manager operator reported degraded because it's operand (daemonset) is not deployed successfully
3) daemonset is not deployed successfuly because k8s did not like that it had duplicate mount entries for the same path:  invalid: spec.template.spec.containers[0].volumeMounts[4].mountPath: Invalid value: \"/etc/pki/ca-trust/extracted/pem\": must be unique\nWorkloadDegraded:

So the next step is to figure out if something else in the operator code is adding the same mount path to the daemonset.

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -295,5 +299,125 @@ func manageOpenShiftControllerManagerDeployment_v311_00_to_latest(client appscli
...
+				corev1.Volume{
+					Name: "global-proxy-ca",
+					VolumeSource: corev1.VolumeSource{
+						ConfigMap: cmSource,
+					},
+				},
+			)
+			mount := corev1.VolumeMount{
+				Name:      "global-proxy-ca",
+				MountPath: "/etc/pki/ca-trust/extracted/pem",
```
bparees: this is the other place you're adding the volume mount.

bindata/v3.11.0/openshift-controller-manager/ds.yaml
gabemontero: sorry my "installer does not apper to like .." was my refraining from spelling out all the detail you just listed and just trying to short hand it somehow

bottom line, yes, need to figure out what is up ... I concur your next step is on point

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
gabemontero: there you go  thanks

gabemontero:
tests are green @bparees 

latest round of changes from your comments pushed

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -295,5 +299,87 @@ func manageOpenShiftControllerManagerDeployment_v311_00_to_latest(client appscli
 	}
 	required.Annotations[util.VersionAnnotation] = os.Getenv("RELEASE_VERSION")
 
+	proxyCfg, err := proxyLister.Get("cluster")
```
adambkaplan: We should have error handling logic here:

1. If `err` is a `NotFound` error, consider the proxy env vars unset, proceed with logic below.
2. If `err` something else, bubble out (should trigger a re-queue).

pkg/operator/sync_openshiftcontrollermanager_v311_00_test.go
```diff
@@ -201,3 +208,93 @@ func TestProgressingCondition(t *testing.T) {
...
+		},
+	)
+	dsClient := kubeClient.AppsV1()
+	indexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
+	proxyConfig := &configv1.Proxy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "cluster",
+		},
+		Spec: configv1.ProxySpec{
+			NoProxy:    "https://no-proxy",
```
adambkaplan: nit - `NO_PROXY` is a comma-separated list of host names. For a unit test this doesn't really matter.

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
gabemontero: So I definitely will update with add a recoreder event and bubbling out / retriggering more immediately - thanks.

On the not found piece, I did some experiments, and
a) the instance is there on a standard install
b) you are not allowed to delete it, even with system:admin, from an install:

```
gmontero ~/go/src/github.com/openshift/cluster-openshift-controller-manager-operator  (ocm-proxy-envs)$ oc get proxies.config.openshift.io
NAME      AGE
cluster   83m
gmontero ~/go/src/github.com/openshift/cluster-openshift-controller-manager-operator  (ocm-proxy-envs)$ oc delete  proxies.config.openshift.io cluster 
Error from server (Forbidden): proxies.config.openshift.io "cluster" is forbidden: deleting required proxies.config.openshift.io resource, named cluster, is not allowed
gmontero ~/go/src/github.com/openshift/cluster-openshift-controller-manager-operator  (ocm-proxy-envs)$ 
```

I'm inclined to minimally record an event if we unexpectedly see not found.

I am afraid that returning the errror and retriggering could create a vicious cycle/loop if the cluster is actually broken, so think will proceed as is as noted.

But further opinions of course welcomed.

gabemontero: update pushed in separate commit @adambkaplan

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -295,5 +299,94 @@ func manageOpenShiftControllerManagerDeployment_v311_00_to_latest(client appscli
...
+					continue
+				}
+				if name == "NO_PROXY" {
+					if !noProxySet {
+						// if no longer set in config, do not add to new list (i.e. delete), force rollout
+						forceRollout = true
+						continue
+					}
+					// note already exists, re-add to new list, with up to date value
+					hasNoProxy = true
```
soltysh: This asks to be a switch, not a set of ifs.

```go
switch name {
case "NO_PROXY":
    ...
case "HTTPS_PROXY":
    ...
case "HTTP_PROXY":
    ...
default:
   newEnvs = append(newEnvs, env)
}

File: pkg/operator/sync_openshiftcontrollermanager_v311_00_test.go (hunk 8 above)
Comment (gabemontero) [reply pkg/operator/sync_openshiftcontrollermanager_v311_00_test.go]: fixed this too

File: pkg/operator/sync_openshiftcontrollermanager_v311_00.go (hunk 9 above)
Comment (gabemontero) [reply pkg/operator/sync_openshiftcontrollermanager_v311_00.go]: update pushed in separate commit thanks @soltysh

**Comment by adambkaplan** [comment]
@gabemontero looks good to me, squash commits and we can make this official.

**Comment by gabemontero** [comment]
10-4 @adambkaplan @soltysh  thanks ..

**Comment by gabemontero** [comment]
... and commits squashed/pushed @adambkaplan @soltysh 

and thanks @bparees for all the help

**Comment by gabemontero** [comment]
`error: could not run steps: step e2e-aws failed: failed to acquire lease: status 503 Service Unavailable, status code 503`

`error: could not run steps: step e2e-aws-operator failed: failed to acquire lease: status 503 Service Unavailable, status code 503`

`error: could not run steps: step e2e-aws-upgrade failed: failed to acquire lease: status 503 Service Unavailable, status code 503`

/retest

**Comment by gabemontero** [comment]
green tests @adambkaplan @soltysh @bparees 

can somebody lgtm ?

File: pkg/operator/configobservation/configobservercontroller/observe_config_controller.go [hunk 10]
```diff
@@ -34,11 +34,13 @@ func NewConfigObserver(
 			configobservation.Listers{
 				ImageConfigLister: configInformers.Config().V1().Images().Lister(),
 				BuildConfigLister: configInformers.Config().V1().Builds().Lister(),
+				ProxyLister:       configInformers.Config().V1().Proxies().Lister(),
```
sttts: shouldn't we install event handlers to notice changes and update?

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -295,5 +299,93 @@ func manageOpenShiftControllerManagerDeployment_v311_00_to_latest(client appscli
...
 
+	proxyCfg, err := proxyLister.Get("cluster")
+	if err != nil {
+		recorder.Eventf("ProxyConfigGetFailed", "Error retrieving global proxy config: %s", err.Error())
+		if !apierrors.IsNotFound(err) {
+			// return daemonset since it is still referenced by caller even with errors
+			return required, false, err
+		}
+	}
+	if err == nil {
```
sttts: `else {`

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -295,5 +299,93 @@ func manageOpenShiftControllerManagerDeployment_v311_00_to_latest(client appscli
...
+
+			if noProxySet && !hasNoProxy {
+				newEnvs = append(newEnvs, corev1.EnvVar{Name: "NO_PROXY", Value: proxyCfg.Status.NoProxy})
+			}
+			if httpProxySet && !hasHTTPProxy {
+				newEnvs = append(newEnvs, corev1.EnvVar{Name: "HTTP_PROXY", Value: proxyCfg.Status.HTTPProxy})
+			}
+			if httpsProxySet && !hasHTTPSProxy {
+				newEnvs = append(newEnvs, corev1.EnvVar{Name: "HTTPS_PROXY", Value: proxyCfg.Status.HTTPSProxy})
+			}
```
sttts: How about this?

```golang
newEnvs := []corev1.EnvVar{}
for _, env := range c.Env {
	name := strings.TrimSpace(env.Name)
	switch name {
	case "HTTPS_PROXY":
		if len(proxyCfg.Status.HTTPSProxy) == 0 {
			continue
		}
		env.Value = proxyCfg.Status.HTTPSProxy
	case ...:
	case ....:
	}
	newEnvs = append(newEnvs, env)
}
forceRollout = forceRollout || !reflect.DeepEqual(newEnvs, c.Env)
```

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
gabemontero: you are absolutely right @sttts 

though this particular piece is a leftover from the confusion over when to use observe_config_controller.go vs. operator.go and its calls to the sync_*.go file

so I think we need
a) cleanup and full revert of the changes here in observe_config_controller.go
b) and install the event handlers as you say around here:  https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/9c985c9efac3916ecaacfcecbce5c317fa8d3cfb/pkg/operator/operator.go#L52-L72

thanks ... getting started

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
gabemontero: I had to tweak it slightly to deal with an empty `c.Env` but I agree your form is better ... will be pushing new commit soon

gabemontero:
updates pushed @stts thanks again

gabemontero:
broad failures across all suites waiting on expected SAs in test namespaces... as an example:

```
Mar 26 16:37:05.309: INFO: Creating project "e2e-test-s2i-build-quota-btfrh"
Mar 26 16:37:05.725: INFO: Waiting on permissions in project "e2e-test-s2i-build-quota-btfrh" ...
Mar 26 16:37:05.752: INFO: Waiting for ServiceAccount "default" to be provisioned...
Mar 26 16:39:03.083: INFO: Waiting for ServiceAccount "deployer" to be provisioned...
[AfterEach] [sig-builds][Feature:Builds] s2i build with a quota
```

The CVO has a nasty looking extra condition on it:

```
                   {
                        "lastTransitionTime": "2020-03-26T15:39:58Z",
                        "message": "Unable to retrieve available updates: currently installed version 0.0.1-2020-03-26-153053 not found in the \"stable-4.5\" channel",
                        "reason": "VersionNotFound",
                        "status": "False",
                        "type": "RetrievedUpdates"
                    }
```

punting

/test e2e-aws

gabemontero:
I see a lot of throttling notifications trying to access the api server in the OCM-O pod logs around the times of the e2e-aws-operator failures ... my assumption is that messed with those e2e's but am getting up to speed on those tests, looking around the artifacts some more before retesting that one.

gabemontero:
> I see a lot of throttling notifications trying to access the api server in the OCM-O pod logs around the times of the e2e-aws-operator failures ... my assumption is that messed with those e2e's but am getting up to speed on those tests, looking around the artifacts some more before retesting that one.

Yeah even the basic, very first test that just reads the OCM clusteroperator suffered a TO, but if you look at the artifacts, that object showed everything was well, and was last updated before the test started.

/test e2e-aws-operator

gabemontero:
pushed fix for e2e-aws-operator that should better account for a false positive we were getting with the `reflect.DeepEquals` change from the last iteration

gabemontero:
CI hiccup across most of the board

/retest

gabemontero:
green tests 

bump - any of the previous reviewers ready to LGTM ?

PR #148, opened 2020-04-06
Bug 1813233: Bumping github.com/openshift/library-go

ricardomaraschini:
By bumping library-go we now bind to "tcp" network type instead of
"tcp4" as before.

This is needed in order to correctly bind our http servers on ipv6
clusters.


ricardomaraschini:
/hold

Until I discovered how to better tackle this.

ricardomaraschini:
No ipv6 support on 4.3 so this is not needed.

PR #159, opened 2020-06-25
Add ForcePull to BuildOverrides

coreydaley:
Adding ForcePull observation for BuildOverrides

https://issues.redhat.com/browse/BUILD-96


pkg/operator/configobservation/util.go
```diff
@@ -21,6 +21,11 @@ func ObserveField(observedConfig map[string]interface{}, val interface{}, fieldN
 	switch v := val.(type) {
 	case int64, bool:
 		err = unstructured.SetNestedField(observedConfig, v, nestedFields...)
+	case *bool:
+		if skipIfEmpty && val == nil {
+			return nil
+		}
+		err = unstructured.SetNestedField(observedConfig, *v, nestedFields...)
```
adambkaplan, line 28: @coreydaley being a bit paranoid. If a caller provides `skipIfEmpty=false` and `val=nil`, we risk a nil pointer dereference panic, no?

I think our best option is to throw an error:

```
if !skipEmpty && val == nil {
    return fmt.Errorf("cannot observe nil value for bool pointer field %s", fieldName)
}
```

coreydaley: updated in current push

adambkaplan:
/lgtm

Generated build pod has `forcePull: true` in the `BUILD` environment variable that is injected into all pods.
@coreydaley next step is to add an e2e test for this feature. We have a test suite in origin which alters the cluster configuration [1], test should set the build override and check that all build pods force pull.

[1] https://github.com/openshift/origin/blob/master/test/extended/builds/cluster_config.go#L344

PR #161, opened 2020-07-06
Bug 1854133: ensure progressing is not set to false if version is missing

gabemontero:
Manual pick to 4.3 since bot failed ... see https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/160#issuecomment-654240158


gabemontero:
code diffs are in line with the original PRs in latter releases ... and unit test is already clean

/assign @adambkaplan

PR #162, opened 2020-07-07
Bug 1826183: Sync proxy CA to openshift-controller-manager

adambkaplan:
Previously, builds mounted the global CA trust bundle as the PEM-encoded TLS
trust bundle. This trust bundle is generated by the Network Operator, which
validates any user-provided CA trust bundle and merges it with the default trus
bundle provided by RHCOS. Using the global CA is problematic because maven and
other JVM-based processes establish trust using `keytool` formatted trust
bundles.

To generate trust bundles in the `keytool` format (as well as other formats used
by RHEL), builds will run `update-ca-trust extract` on startup. To merge the
trust bundle correctly, the build controller needs to mount any user-provided CA
in a well-known location. The openshift-builder process is then responsible for
copying the CA to /etc/ca-trust/source/anchors prior to running
`update-ca-trust extract`.

To achieve this, the operator is being enhanced with the following:

* Added a separate controller for observing the user CA config.
* Added a ResourceSync controller to copy the user CA to the
  openshift-controller-manager namespace.
* Updated starter to launch the new controllers.

The operator will continue to create the `openshift-global-ca` ConfigMap to
ensure that the openshift controller manager does not have an outage during
upgrades. This can be removed in a subsequent release.

See also openshift/builder#153, openshift/builder#158, openshift/openshift-controller-manager#119


pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -253,6 +254,10 @@ func manageOpenShiftGlobalCAConfigMap_v311_00_to_latest(kubeClient kubernetes.In
 	existing, err := client.ConfigMaps(util.TargetNamespace).Get(context.TODO(), "openshift-global-ca", metav1.GetOptions{})
 	// Ensure we create the ConfigMap for the global CA, and that it has the right labels
 	// Lifted from library-go for the most part
+
+	// Bug 1826183: The global CA bundle is not intended to be used with workloads which run `update-ca-trust extract` on their own.
+	// In 4.6 we need to create this ConfigMap with the trust bundle injected so that the controller manager does not have an outage during upgrade.
+	// In 4.7 and beyond this ConfigMap should be deleted.
```
gabemontero, line 260: So to connect some dots:
- builds, either docker or s2i strategy, *always* call `update-ca-trust extract` now?
- if yes, then some clarification here (vs. explanations I assume exist in openshift/builder or openshift/openshift-controller-manager) on how the global CAs are picked up, so things work with the proxy in addition to the direct contact to external repos like nexus, artifactory, etc. would be good
- if no, then don't we still need global CA bundle for the "non update-ca-trust" scenarios

pkg/operator/usercaobservation/user_ca_observation_controller.go
```diff
@@ -0,0 +1,94 @@
...
+
+	source, err := c.findProxyCASource()
+	if err != nil {
+		return err
+	}
+	destination := resourcesynccontroller.ResourceLocation{
+		Namespace: util.TargetNamespace,
+		Name:      "openshift-user-ca",
+	}
+	err = c.resourceSyncer.SyncConfigMap(destination, source)
```
gabemontero, line 74: so either user ca is a singleton like proxy ca, or SyncConfigMap is doing a merge of some sort 

come clarification either way would be good

also, some details on why we need to copy ... i.e. the specfics around what breaks down with update-ca-trust if we read the "original" global CA vs. the user CA we've copied to would be good

How does the fact that we copy the proxy's trusted CA reconcile, get us out of jail, play nice, etc. with what I read in https://github.com/openshift/api/blob/master/config/v1/types_proxy.go#L44-L69

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
adambkaplan: >     * builds, either docker or s2i strategy, _always_ call `update-ca-trust extract` now?

Yes - this was added in https://github.com/openshift/builder/pull/158

>     * if yes, then some clarification here (vs. explanations I assume exist in openshift/builder or openshift/openshift-controller-manager) on how the global CAs are picked up, so things work with the proxy in addition to the direct contact to external repos like nexus, artifactory, etc. would be good

The new flow will be a follows:

1. ocm-o copies the user trusted CA to `openshift-controller-manager/openshift-user-ca`
2. For each build, openshift-controller-manager reads `openshift-user-ca` and copies that into a `ConfigMap` in the build's namespace, mounting it at `/var/run/configs/openshift.io/pki/tls-ca-bundle.pem` inside the build pod containers.
3. When builds run, it copies this CA to `/etc/pki/ca-trust/source/anchors` and runs `update-ca-trust extract`. This merges the custom CA with the default CAs provided by the builder image.

I will update with additional comments.

pkg/operator/usercaobservation/user_ca_observation_controller.go
adambkaplan: I updated the comment above to clarify what is going on overall, and why we need to copy the user CA to the openshift-controller-manager namespace. We get out of jail because build pods run `update-ca-trust extract` themselves.

My updated commit message indicates why we need to do this (namely, the issue with `keytool` formatted trust bundles).

PR #165, opened 2020-07-29
Bug 1809892: "unmanaged" causes unstable OCM pods.

otaviof:
This PR resolves [bug-1809892](https://bugzilla.redhat.com/show_bug.cgi?id=1809892) by adding a short-circuit in the reconciliation loop (`sync`) on which `managementState` options `Removed` and `Unmanaged` are marked as unsupported. No further actions to reconcile previously managed objects are taken.

## Testing

Edit cluster operator configuration `openshiftcontrollermanagers` and set it as `Unmanaged`:

```sh
oc patch openshiftcontrollermanagers.operator.openshift.io/cluster --namespace="openshift-controller-manager-operator" --patch '{"spec":{"managementState": "Unmanaged"}}' --type=merge
```

The pods in `openshift-controller-manager` namespace are intact.

```sh
oc --namespace=openshift-controller-manager get pods --watch
```

Finally, by setting up as `Managed`, the `controller-manager` pods will be restarted a single time:

```sh
oc patch openshiftcontrollermanagers.operator.openshift.io/cluster --namespace="openshift-controller-manager-operator" --patch '{"spec":{"managementState": "Managed"}}' --type=merge
```

### DIFF
--- a/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
+++ b/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	appsclientv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
 	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/klog"
 
 	operatorapiv1 "github.com/openshift/api/operator/v1"
 
@@ -90,7 +91,14 @@ func syncOpenShiftControllerManager_v311_00_to_latest(c OpenShiftControllerManag
 		errors = append(errors, fmt.Errorf("%q: %v", "openshift-global-ca", err))
 	}
 
-	forceRollout = forceRollout || operatorConfig.ObjectMeta.Generation != operatorConfig.Status.ObservedGeneration
+	klog.Infof("Modified resources: clientCA=%v, serviceCA=%v, globalCA=%v", clientCAModified, serviceCAModified, globalCAModified)
+	klog.Infof("Observed generation '%d', current configuration generation '%d'", operatorConfig.Status.ObservedGeneration, operatorConfig.ObjectMeta.Generation)
+
+	// Bug 1809892: setting force-rollout only when management status is "managed", to avoid
+	// comparing observed generation with current. When "unmanaged", the observed generation is not
+	// incremented, while changing management status does.
+	isManaged := operatorConfig.Spec.ManagementState == operatorapiv1.Managed
+	forceRollout = forceRollout || (isManaged && operatorConfig.ObjectMeta.Generation != operatorConfig.Status.ObservedGeneration)
 	forceRollout = forceRollout || configMapModified || clientCAModified || serviceCAModified || globalCAModified
 
 	// our configmaps and secrets are in order, now it is time to create the DS


pkg/operator/operator.go
```diff
@@ -85,46 +85,36 @@ func NewOpenShiftControllerManagerOperator(
...
-		condition.Reason = reasonString
-		condition.Message = messageString
-		v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, condition)
-	}
-	if !equality.Semantic.DeepEqual(operatorConfig.Status, originalOperatorConfig.Status) {
-		if _, err := c.operatorConfigClient.OpenShiftControllerManagers().UpdateStatus(context.TODO(), operatorConfig, metav1.UpdateOptions{}); err != nil {
-			return err
+		forceRequeue, err := syncOpenShiftControllerManager_v311_00_to_latest(c, operatorConfig)
+		if forceRequeue && err != nil {
+			c.queue.AddRateLimited(workQueueKey)
```
gabemontero, line 113: why would we not empty out the conditions when managed and we did not get an error ?

otaviof: In the original changes, we were getting this error back:
```
E0730 08:02:46.640390       1 operator.go:164] key failed with : Operation cannot be fulfilled on openshiftcontrollermanagers.operator.openshift.io "cluster": the object has been modified; please apply your changes to the latest version and try again
```

With the changes in this PR the error is gone, and we add the following condition:

```
$ oc --namespace="openshift-controller-manager-operator" get openshiftcontrollermanagers.operator.openshift.io cluster -o json |jq '.status.conditions[-1]'
{
  "lastTransitionTime": "2020-07-30T08:27:41Z",
  "message": "the controller manager spec is set to Unmanaged state",
  "reason": "UnmanagedUnsupported",
  "status": "True",
  "type": "WorkloadDegraded"
}
```

gabemontero: what happens when you switch from unmanaged back to managed?

otaviof: @gabemontero I misunderstood the use of status conditions in the operator, thanks for pointing this out. In the latest commit I'm reverting to the original behavior, please consider.

pkg/operator/operator.go
```diff
@@ -85,46 +85,43 @@ func NewOpenShiftControllerManagerOperator(
...
-		condition.Reason = reasonString
-		condition.Message = messageString
-		v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, condition)
-	}
-	if !equality.Semantic.DeepEqual(operatorConfig.Status, originalOperatorConfig.Status) {
-		if _, err := c.operatorConfigClient.OpenShiftControllerManagers().UpdateStatus(context.TODO(), operatorConfig, metav1.UpdateOptions{}); err != nil {
-			return err
+		// when managed, all resources under control of this operator will be
+		// modified to reflect desired state.
+		forceRequeue, err := syncOpenShiftControllerManager_v311_00_to_latest(c, operatorConfig)
```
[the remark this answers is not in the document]
otaviof: I'm using the following snippet to investigate the conditions, and how they are changing depending on the management state.

```sh
oc get openshiftcontrollermanagers.operator.openshift.io cluster \
  --namespace="openshift-controller-manager-operator" \
  --output=json \
    |jq '.status.conditions[]'
```
### Unmanaged State

When the configuration is set to `unmanaged` state, the conditions are rewritten as:

```json
{
  "lastTransitionTime": "2020-07-12T05:13:37Z",
  "status": "False",
  "type": "ConfigObservationDegraded"
}
{
  "lastTransitionTime": "2020-08-01T08:42:44Z",
  "message": "the controller manager spec was set to Unmanaged state, but that is unsupported, and has no effect on this condition",
  "reason": "UnmanagedUnsupported",
  "status": "True",
  "type": "Available"
}
{
  "lastTransitionTime": "2020-08-01T09:04:15Z",
  "message": "the controller manager spec was set to Unmanaged state, but that is unsupported, and has no effect on this condition",
  "reason": "UnmanagedUnsupported",
  "status": "False",
  "type": "Progressing"
}
{
  "lastTransitionTime": "2020-07-12T05:13:54Z",
  "message": "the controller manager spec was set to Unmanaged state, but that is unsupported, and has no effect on this condition",
  "reason": "UnmanagedUnsupported",
  "status": "False",
  "type": "WorkloadDegraded"
}
```

### Managed State

On the default, `managed` state, the conditions are:

```json
{
  "lastTransitionTime": "2020-07-29T04:50:25Z",
  "status": "False",
  "type": "ConfigObservationDegraded"
}
{
  "lastTransitionTime": "2020-08-05T14:03:40Z",
  "status": "True",
  "type": "Available"
}
{
  "lastTransitionTime": "2020-08-05T14:03:40Z",
  "status": "False",
  "type": "Progressing"
}
{
  "lastTransitionTime": "2020-07-29T04:50:46Z",
  "status": "False",
  "type": "WorkloadDegraded"
}
```

Considering that all conditions are being mutated, @gabemontero, I think adding one more condition here, it would be lost among all others.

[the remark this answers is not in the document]
gabemontero: I'm not following what you mean by `I think adding one more condition here, it would be lost among all others.` @otaviof 

I'm going try to re-frame things a bit ... let's go with some test-driven development:

write both a unit test and a e2e at https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/test/e2e/cluster_openshift_controller_manager_operator_test.go that 
- validates going managed, unmanaged, and then managed again, where the conditions are checked at each stage to confirm what you see above (I bet the unit test can take care of this one)
- validates the problem noted in the bug, that the daemonset does not continually rollout when either unmanaged or managed (this one most likely can only be validated in the e2e test with an actual operator running)

whatever code change makes those new tests work will be what we go with, minus any stylistic discussion that might ensue

sttts:
/hold

I don't like operator specific solutions for this. The management state is something generic. We should have generic behaviour and a generic implementation in library-go.

mfojtik:
https://github.com/openshift/library-go/tree/master/pkg/operator/management

otaviof:
> I don't like operator specific solutions for this. The management state is something generic. We should have generic behaviour and a generic implementation in library-go.

@sttts, @mfojtik, thanks for the pointer to library-go, was educative exploring it.

I'm pushing another set of changes to address the trigger of the [bug-1809892](https://bugzilla.redhat.com/show_bug.cgi?id=1809892), at the same level the operator probes for observed vs. current generation. Please consider.

otaviof:
We've spoken about this PR today, and we decided to close it, and pursue a more extensive refactoring later on (for OS 4.7).

PR #167, opened 2020-08-03
Bug 1855640: Rebase to kube-0.19.0-rc.3

coreydaley:


coreydaley:
/assign @adambkaplan 
fyi @gabemontero @otaviof

pkg/operator/generic_client.go
```diff
@@ -38,6 +39,10 @@ func (p *genericClient) GetOperatorState() (*operatorapiv1.OperatorSpec, *operat
 	return &instance.Spec.OperatorSpec, &instance.Status.OperatorStatus, instance.ResourceVersion, nil
 }
 
+func (p *genericClient) GetObjectMeta() (*metav1.ObjectMeta, error) {
```
coreydaley, line 42: Had to add this function to be in compliance with being an OperatorClient

adambkaplan:
This needs deeper investigation - most builds are failing with:

```
Timed out waiting for Openshift Controller Manager to be rolled out with updated internal registry hostname
occurred
```

This implies that the operator is reporting `Progressing: True` when either a) it shouldn't, or b) the controller manager is crashlooping.

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -382,5 +382,5 @@ func manageOpenShiftControllerManagerDeployment_v311_00_to_latest(client appscli
 		}
 	}
 
-	return resourceapply.ApplyDaemonSet(client, recorder, required, resourcemerge.ExpectedDaemonSetGeneration(required, generationStatus), forceRollout)
+	return resourceapply.ApplyDaemonSet(client, recorder, required, resourcemerge.ExpectedDaemonSetGeneration(required, generationStatus))
```
adambkaplan, line 385: What is the impact of not having the `forceRollout` option? I'm pretty sure we heavily relied on this in the operator.

pkg/operator/generic_client.go
adambkaplan: Perhaps throwing an error here is bad?

coreydaley: There is precedent in another repo (i can find it if you want me to) for returning the not implemented error here.

adambkaplan: Perhaps...but if we have a canonical instance of our object, and it has standard metadata, why not return the real thing?

coreydaley: The struct for this has informers and a client, there is no metadata?

coreydaley:
@adambkaplan tests are green and comments have been responded to

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
coreydaley: Fixed, looks like the ApplyDaemonsetForce is going away in a bit so we'll have to fix that at some point also.

adambkaplan: Personally I'd like to do a major refactor of this operator. `library-go` has a lot of canned controllers that are shared across multiple CVO operators that we should take advantage of.

adambkaplan: Tracking the refactor in https://issues.redhat.com/browse/BUILD-113

pkg/operator/generic_client.go
adambkaplan: Looking at the other methods that implement the interface, it seems that the Get/Update methods fetch the canonical operator config record from the lister:

```
resource, err := p.informers.Operator().V1().OpenShiftControllerManagers().Lister().Get("cluster")
...
```

The new method was added in https://github.com/openshift/library-go/pull/817. It appears this method is only used by some operators today, but not ours at present. IMO the implementation here is straightforward, so we should do it now vs. dealing with a bug down the road.

coreydaley: updated in current push

PR #169, opened 2020-08-18
Bug 1820075: Allow ingress-to-route to update ingress status

Miciah:
Add a new `system:openshift:openshift-controller-manager:ingress-to-route-controller` cluster role and cluster role binding for the ingress-to-route controller. The new cluster role is based on the `system:openshift:controller:ingress-to-route-controller` cluster role from the bootstrap policy, with the following changes:

* The new role references the `ingresses` resource in the `networking.k8s.io` group, instead of using the `ingresses` resource in the `extensions` group, which is deprecated.

* The new role allow updating the `ingresses/status` subresource.

* `bindata/v3.11.0/openshift-controller-manager/ingress-to-route-controller-clusterrole.yaml`: New file, based on `system:openshift:controller:ingress-to-route-controller` from the bootstrap cluster roles.  Replace the `extensions` API group with the `networking.k8s.io` API group for `ingresses`.  Allow the ingress-to-route controller to update the status of `ingresses`.
* `bindata/v3.11.0/openshift-controller-manager/ingress-to-route-controller-clusterrolebinding.yaml`: New file, based on `system:openshift:controller:ingress-to-route-controller` from the bootstrap cluster role bindings.
* `pkg/operator/sync_openshiftcontrollermanager_v311_00.go` (`syncOpenShiftControllerManager_v311_00_to_latest`): Add the new resources.
* `pkg/operator/v311_00_assets/bindata.go`: Regenerate.

----

Related to https://github.com/openshift/openshift-apiserver/pull/129 and https://github.com/openshift/openshift-controller-manager/pull/130.

### DIFF
--- a/bindata/v3.11.0/openshift-controller-manager/ingress-to-route-controller-clusterrolebinding.yaml
+++ b/bindata/v3.11.0/openshift-controller-manager/ingress-to-route-controller-clusterrolebinding.yaml
@@ -1,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: system:openshift:openshift-controller-manager:ingress-to-route-controller
+roleRef:
+  kind: ClusterRole
+  name: system:openshift:openshift-controller-manager:ingress-to-route-controller
+subjects:
+- kind: ServiceAccount
+  namespace: openshift-infra
+  name: ingress-to-route-controller
--- a/bindata/v3.11.0/openshift-controller-manager/ingress-to-route-controller-clusterrole.yaml
+++ b/bindata/v3.11.0/openshift-controller-manager/ingress-to-route-controller-clusterrole.yaml
@@ -1,0 +1,55 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:openshift:openshift-controller-manager:ingress-to-route-controller
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  - services
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - networking.k8s.io
+  resources:
+  - ingress
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - networking.k8s.io
+  resources:
+  - ingresses/status
+  verbs:
+  - update
+- apiGroups:
+  - route.openshift.io
+  resources:
+  - routes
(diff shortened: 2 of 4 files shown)


bindata/v3.11.0/openshift-controller-manager/ingress-to-route-controller-clusterrolebinding.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ClusterRoleBinding
+metadata:
+  name: system:openshift:openshift-controller-manager:ingress-to-route-controller
+roleRef:
+  kind: ClusterRole
+  name: system:openshift:openshift-controller-manager:ingress-to-route-controller
+subjects:
+- kind: ServiceAccount
+  namespace: openshift-infra
+  name: ingress-to-route-controller
```
adambkaplan, line 11: @Miciah why is the ocm-o managing the `ingress-to-route-controller` service account in the `openshift-infra` namespace?

adambkaplan (commented):
Commits should be squashed before merging.

I'd like to know why ocm-o should manage/alter the service account roles. This SA isn't something that ocm-o declares in its CVO manifests.

bindata/v3.11.0/openshift-controller-manager/ingress-to-route-controller-clusterrolebinding.yaml
Miciah: My memory is fuzzy, but I think I followed the lead of `separate-sa-role.yaml`.  What namespace should it use?

adambkaplan: Some poking of the cluster and code led me to figure out why this is necessary:

1. Each controller has its own set of RBAC roles/permissions. To accomplish this, these controllers run as a service account in the `openshift-infra` namespace.
2. Therefore, it's appropriate for OCM-O to manage these cluster roles.

Note @gabemontero we will need to do something similar if we want to narrow the build controller's permissions.

adambkaplan (approved):
@Miciah looks good now that I've done some poking.

I don't think we need separate commits - you can go ahead and squash into one commit, and I can add the lgtm

/approve

adambkaplan:
fyi @otaviof @coreydaley - similar steps will be needed if we want to prevent the build controller from creating privileged containers.

Miciah:
Latest push squashes commits.

adambkaplan (approved):
/lgtm

PR #173, opened 2020-09-07
Bug 1809892: Workload Refactoring

otaviof:
Initial refactoring of operator code base, in order to create the path to, on a near future PR, move common business logic towards [`library-go`](https://github.com/openshift/library-go).

This pull-request introduce `workload` package, responsible for the reconciliation of the resources managed by the operator, while `operator.go` handles the common Kubernetes API logic. 

The reconciliation (sync) is executed in the new `workload` package, where the logic is re-organized and documented, plus the error handling is following Go 1.13 standards.

## Performance

The cluster configuration is now handled by informer/listers and therefore we take advantage of the split-client behavior in Kubernetes. That avoids a few requests against the API-Server each reconciliation loop.

## Testing

Execute test targets via `Makefile`, as per:

```sh
make test
make test-e2e
```

The end-to-end tests are simulating the change of management status, and asserting status conditions on each step. And, unit test is checking how many API calls are executed upon reconciliation and more.

## Bug 180989

This PR fixes [Bug 1809892](https://bugzilla.redhat.com/show_bug.cgi?id=1809892) by allowing managed `DaemonSet` generation skew, the situation was highlighted on #165 previously.


pkg/operator/operator.go
```diff
@@ -19,34 +19,28 @@ import (
 
 	operatorapiv1 "github.com/openshift/api/operator/v1"
 	configinformerv1 "github.com/openshift/client-go/config/informers/externalversions/config/v1"
-	proxyvclient1 "github.com/openshift/client-go/config/listers/config/v1"
 	operatorclientv1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
 	operatorinformersv1 "github.com/openshift/client-go/operator/informers/externalversions/operator/v1"
+	"github.com/openshift/cluster-openshift-controller-manager-operator/pkg/operator/workload"
```
coreydaley, line 24: As a follow up to this pull request, it may be worth running our import sorting script on this repo.

pkg/operator/operator.go
```diff
@@ -56,25 +50,32 @@ func NewOpenShiftControllerManagerOperator(
 	kubeClient kubernetes.Interface,
 	recorder events.Recorder,
 ) *OpenShiftControllerManagerOperator {
+	proxyLister := proxyInformer.Lister()
+	q := workqueue.NewNamedRateLimitingQueue(
```
coreydaley, line 54: I would like to see q and w replaced with something more descriptive.

pkg/operator/operator.go
```diff
@@ -56,25 +50,32 @@ func NewOpenShiftControllerManagerOperator(
...
+		workload:             w,
 	}
 
 	operatorConfigInformer.Informer().AddEventHandler(c.eventHandler())
 	proxyInformer.Informer().AddEventHandler(c.eventHandler())
-	kubeInformersForOpenshiftControllerManager.Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
-	kubeInformersForOpenshiftControllerManager.Core().V1().ServiceAccounts().Informer().AddEventHandler(c.eventHandler())
-	kubeInformersForOpenshiftControllerManager.Core().V1().Services().Informer().AddEventHandler(c.eventHandler())
-	kubeInformersForOpenshiftControllerManager.Apps().V1().Deployments().Informer().AddEventHandler(c.eventHandler())
+	kubeInformersForOpenshiftControllerManager.Core().V1().ConfigMaps().
```
coreydaley, line 67: We don't usually split lines based on a certain character count, I think it makes it harder  to read (we went through this with the image registry/image registry operator also, but I think that was before your time)

pkg/operator/workload/util.go
```diff
@@ -0,0 +1,28 @@
+package workload
```
coreydaley, line 1: I don't think this makes sense, appending a flag to force rollout to an error slice.
It would make more sense to either have a separate variable for the force rollout, or create a data structure that stores a force rollout bool and a slice of errors.

pkg/operator/workload/workload.go
```diff
@@ -0,0 +1,305 @@
...
+
+// Workload reprents the operator components and the rollout approach.
+type Workload struct {
+	targetImagePullSpec string                    // daemon-set image pull-spec
+	proxyLister         proxyclientv1.ProxyLister // proxy client lister
+	kubeClient          kubernetes.Interface      // kubernetes client
+	recorder            events.Recorder           // event recorder
+}
+
+// WorWorkloadDegradedCondition status condition to denote manage workload is degraded.
```
coreydaley, line 28: s/WorWorkloadDegradedCondition/WorkloadDegradedCondition

pkg/operator/operator.go
otaviof: I think the empty line on the import made the sorting done by my editor different than usual. Let me fix it, and see if it will match expected sequence.

pkg/operator/operator.go
otaviof: Sure, no problem.

pkg/operator/operator.go
otaviof: Okay, I understand that can be a polemic subject. I'm revering the changes :-)

pkg/operator/workload/util.go
otaviof: The suggested change makes possible to track which component is triggering the "force rollout" scenario, and by leveraging the slice of errors, it becomes possible to identify the ones will need special handling. We could achieve the same results with a custom type, however, the suggested changes are following a [common golang pattern](https://blog.golang.org/go1.13-errors).

pkg/operator/workload/workload.go
otaviof: Done. Thanks.

test/e2e/util.go
```diff
@@ -0,0 +1,54 @@
...
+
+// assertOperatorConditions compare two slices of operator conditions ignoring timestamps.
+func assertOperatorConditions(t *testing.T, expected, actual []operatorv1.OperatorCondition) bool {
+	for _, e := range expected {
+		matches := false
+		for _, a := range actual {
+			if matches {
+				continue
+			}
+			if e.Type == a.Type && e.Status == a.Status && e.Reason == a.Reason {
```
gabemontero, line 44: Are there never any messages set on these conditions?

test/e2e/cluster_openshift_controller_manager_operator_test.go
```diff
@@ -113,3 +117,77 @@ func TestClusterImageConfigObservation(t *testing.T) {
...
+			Type:   condition.ResourceSyncControllerDegradedConditionType,
+			Status: operatorv1.ConditionFalse,
+		}, {
+			Type:   workload.WorkloadDegradedCondition,
+			Reason: "UnmanagedUnsupported",
+			Status: operatorv1.ConditionFalse,
+		}},
+	}
+
+	// going from managed to unmanaged, and again to original state
```
gabemontero, line 170: Yay :-) !!

pkg/operator/operator.go
```diff
@@ -99,12 +115,17 @@ func (c OpenShiftControllerManagerOperator) sync() error {
...
+				"and has no effect on this condition",
+			string(operatorConfig.Spec.ManagementState),
+		)
+
+		// as we are ignoring / not supporting unmanaged/removed, we still process any other inputs
+		// to the sync/reconciliation unlike what we used to do
 	case operatorapiv1.Managed:
-		// we want to empty out the reason/message string if transitioning from the other phases so the default setting above is good
+		// we want to empty out the reason/message string if transitioning from the other phases so
+		// the default setting above is good
```
gabemontero, line 128: Yeah in relations to my comment in the assertOperatorConditions method, we should validate that the messages field is emptied out when going back to managed

pkg/operator/operator.go
```diff
@@ -84,13 +78,35 @@ func NewOpenShiftControllerManagerOperator(
...
+func (c OpenShiftControllerManagerOperator) updateStatus(
+	original *operatorapiv1.OpenShiftControllerManager,
+	modified *operatorapiv1.OpenShiftControllerManager,
+) (*operatorapiv1.OpenShiftControllerManager, error) {
+	if equality.Semantic.DeepEqual(original.Status, modified.Status) {
+		return modified, nil
+	}
+	opts := metav1.UpdateOptions{}
+	return c.operatorConfigClient.OpenShiftControllerManagers().
+		UpdateStatus(context.TODO(), modified, opts)
```
gabemontero, line 99: why declare opts ... just pass metav1.UpdateOptions{} as a parameter

pkg/operator/operator.go
```diff
@@ -84,13 +78,35 @@ func NewOpenShiftControllerManagerOperator(
 	return c
 }
 
+// getOperatorConfig execute get request using operator config client.
+func (c OpenShiftControllerManagerOperator) getOperatorConfig() (
+	*operatorapiv1.OpenShiftControllerManager, error,
+) {
+	return c.operatorConfigClient.OpenShiftControllerManagers().
```
gabemontero, line 85: since we are already have an operatorConfigInformer set up, we should be using a lister off of that to fetch the config and avoid hits against the api server

a few releases ago Clayton dinged a bunch of operators for for not minimizing there hits against the api server for stuff like this

pkg/operator/workload/manage.go
```diff
@@ -0,0 +1,288 @@
+package workload
```
gabemontero, line 1: The PR description mentions moving https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go to this new package, but I'm a little unclear on what exactly the change gains us.

Can you elaborate on the motivations @otaviof ?  Of if this was more than simply moving code from https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go and I'm missing what was added, can you highlight that either by PR comment or code comment ?

thanks

gabemontero (commented):
A first pass review @otaviof 

Aside from any updates based on my comments, oo you in fact have more changes coming?  Or should the WIP indicator be removed and we go at this in earnest.

pkg/operator/operator.go
```diff
@@ -113,17 +134,16 @@ func (c OpenShiftControllerManagerOperator) sync() error {
 		condition.Message = messageString
 		v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, condition)
 	}
-	if !equality.Semantic.DeepEqual(operatorConfig.Status, originalOperatorConfig.Status) {
-		if _, err := c.operatorConfigClient.OpenShiftControllerManagers().UpdateStatus(context.TODO(), operatorConfig, metav1.UpdateOptions{}); err != nil {
-			return err
-		}
+	if operatorConfig, err = c.updateStatus(originalOperatorConfig, operatorConfig); err != nil {
+		return err
```
adambkaplan, line 138: Worth asking since we are refactoring - do we need potentially two calls to `updateStatus` per sync loop?

pkg/operator/workload/manage.go
```diff
@@ -0,0 +1,288 @@
...
+	events.Recorder,
+	*operatorapiv1.OpenShiftControllerManager,
+) (*corev1.ConfigMap, bool, error)
+
+type manageClientCAFn func(
+	client coreclientv1.CoreV1Interface,
+	recorder events.Recorder,
+) (bool, error)
+
+func manageOpenShiftControllerManagerClientCA_v311_00_to_latest(
```
adambkaplan, line 41: We should drop references to v311 everywhere. This is an artifact of the very early OCP 4 days when direct upgrades from v3.11 to v4.x was on the table.

Note - perhaps should be done in a separate PR.

pkg/operator/operator.go
otaviof: Was just a left over from a previous interaction. In general, that's a way to reduce line width.

pkg/operator/workload/manage.go
otaviof: Since the functions containing `v311` are coming from the original version, I think we should make the renaming in a subsequent PR is a good idea.

test/e2e/util.go
otaviof: Yes, there are messages in these conditions. It use to be snippet repeated around, and it has been refactored to a generic sub-routine.

pkg/operator/operator.go
otaviof: For this specific case, detecting message fields have been emptied out, we will need to recur to inspect which actions have been registered in the (fake) client. We can teak the approach of testing to assert this test-case as well.

In the current implementation we only validate the final state of the resources, and `sync()` updates the cluster config status twice. The status update does not register a event, but we can still obtain this information in the client.

pkg/operator/operator.go
otaviof: As per @gabemontero [comment](https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/173#discussion_r488654426), we do need to empty out the conditions when going from one management state to another.

Note that the `updateStatus()` routine only issues the update when the status is different, so potentially we would only have two API calls issued when changing the management state.

pkg/operator/operator.go
gabemontero: bump on this one @otaviof .... is this on your to-do list?  Or is it in progress but just not pushed yet?

test/e2e/util.go
gabemontero: so are you saying you are validating the messages in another routine @otaviof ?  If so, can you put a url link to it here?

otaviof: https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/dbdac5e7ba188fcd0b35867cc7e9f6ce07e5c66e/pkg/operator/operator_test.go#L144-L159

gabemontero: ah there you go ... thanks @otaviof

pkg/operator/operator.go
otaviof: Please consider latest commit.

adambkaplan:
@otaviof PR needs rebase

pkg/operator/operator.go
adambkaplan: @otaviof looks like this needs to be updated to use a lister cache.

pkg/operator/workload/manage.go
adambkaplan: @otaviof this comment still needs to be addressed.

otaviof: > The PR description mentions moving https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go to this new package, but I'm a little unclear on what exactly the change gains us.

By moving `sync_openshiftcontrollermanager_v311_00.go` to a standalone package we gain isolation of the primary logic of this operator in a single place. And based upon other operators OpenShift organization I see that "workload" is a common name for this type of package.

We should refactor the `workload` package further, possibly adopting  more of library-go's `NewResourceSyncController`, in subsequent PRs.

> Can you elaborate on the motivations @otaviof ? Of if this was more than simply moving code from https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go and I'm missing what was added, can you highlight that either by PR comment or code comment ?

In this PR we've improved the unit and integration tests, plus refactored the routes into a new package. By doing so the code was reorganized into smaller/simpler functions, all doc-comments have been reviewed and such.

What's your take on this, @gabemontero?

pkg/operator/operator.go
otaviof: @adambkaplan it should be present on commit c7a5f0af, am I overlooking something?

pkg/operator/operator.go
sttts: I don't understand why the deepequal was removed. Why call the server if you don't have to?

pkg/operator/workload/manage.go
```diff
@@ -0,0 +1,288 @@
+package workload
+
+import (
```
sttts, line 3: Having code moves and real changes in the same commit is bad. Nobody can review this in a sensible way.

pkg/operator/operator.go
```diff
@@ -87,8 +86,7 @@ func NewOpenShiftControllerManagerOperator(
 func (c OpenShiftControllerManagerOperator) getOperatorConfig() (
 	*operatorapiv1.OpenShiftControllerManager, error,
 ) {
-	return c.operatorConfigClient.OpenShiftControllerManagers().
-		Get(context.TODO(), "cluster", metav1.GetOptions{})
+	return c.operatorConfigLister.Get("cluster")
```
sttts, line 89: the only actual change in an innocently looking commit.

sttts:
Am not so sure I am happy with what I am seeing here. It is a considerable rewrite / restructuring of code and it makes ocmo to look differently than the other control-plane operators.

It feels like this PR addresses much more than just https://bugzilla.redhat.com/show_bug.cgi?id=1809892.

mfojtik:
I think a refactoring I would appreciate first would be to refactor the controllers to use library-go factory, which will align it with rest of the control plane operators AND it will fix the graceful termination problem this operator still have.

pkg/operator/workload/manage.go
otaviof: Indeed, I need to redo the commits on this branch and make sure the code move is addressed.

pkg/operator/operator.go
otaviof: It does deserve a highlight on the PR description, sorry. It end up being part of this PR via the review process, on @gabemontero's comment.

otaviof:
> Am not so sure I am happy with what I am seeing here. It is a considerable rewrite / restructuring of code and it makes ocmo to look differently than the other control-plane operators.

The overall refactoring attempt was to make it get closer from other control-plane operators, however, admittedly it's the first time I'm trying so, and it will need more effort in order to get there. 

> It feels like this PR addresses much more than just https://bugzilla.redhat.com/show_bug.cgi?id=1809892.

Correct, it does more than addressing the bug. Please consider internal card, [BUILD-113](https://issues.redhat.com/browse/BUILD-113).

pkg/operator/operator.go
otaviof: @sttts the `DeepEqual` call has been moved here [operator.go:93](https://github.com/otaviof/cluster-openshift-controller-manager-operator/blob/83e1703c7acda4d705d108ca9f38a1a24ec2d0bd/pkg/operator/operator.go#L93).

PR #176, opened 2020-09-16
Bug 1879253: Remove .rhel7 Dockerfile

adambkaplan:
The rhel7 Dockerfile variant is not needed now that builds are based on ubi8 and the primary Dockerfile data is synced with ocp-build-data.


soltysh (approved):
/lgtm

PR #190, opened 2020-12-16
WIP - Bug 1899400: Add liveness check for ocm-o

adambkaplan:
Add liveness health checks for openshift-controller-manager-operator.


adambkaplan:
Wow, this bombed out even worse than I expected. OCM-O and its operand are not present in must-gather. I'm pretty sure the server is already initialized, therefore we can't add a health check after the fact.

adambkaplan:
/close

This needs far more work within library-go

PR #197, opened 2021-02-04
Bug 1911470: Set registry routes in operand config

ricardomaraschini:
This PR sets config.RegistryURLs to the Image Registry exposed routes.

### DIFF
--- a/pkg/operator/configobservation/images/observe_images.go
+++ b/pkg/operator/configobservation/images/observe_images.go
@@ -52,3 +52,50 @@ func ObserveInternalRegistryHostname(genericListers configobserver.Listers, reco
 
 	return observedConfig, errs
 }
+
+// ObserveExternalRegistryHostnames observers information about registry external URLs,
+// aka Routes. It retrieves this information from cluster's Image config.
+func ObserveExternalRegistryHostnames(
+	genericListers configobserver.Listers,
+	recorder events.Recorder,
+	existingConfig map[string]interface{},
+) (map[string]interface{}, []error) {
+	var errs []error
+	prevObservedConfig := map[string]interface{}{}
+	listers := genericListers.(configobservation.Listers)
+
+	// first observe all the existing config values so that if we get any errors
+	// we can at least return those.
+	cfgpath := []string{"dockerPullSecret", "registryURLs"}
+	currentURLs, _, err := unstructured.NestedStringSlice(existingConfig, cfgpath...)
+	if err != nil {
+		return prevObservedConfig, append(errs, err)
+	}
+	if len(currentURLs) > 0 {
+		if err := unstructured.SetNestedStringSlice(
+			prevObservedConfig, currentURLs, cfgpath...,
+		); err != nil {
+			return prevObservedConfig, append(errs, err)
+		}
+	}
+
+	observedConfig := map[string]interface{}{}
+	configImage, err := listers.ImageConfigLister.Get("cluster")
+	if errors.IsNotFound(err) {
+		klog.V(2).Infof("images.config.openshift.io/cluster: not found")
+		return observedConfig, errs
+	} else if err != nil {
+		return prevObservedConfig, append(errs, err)
+	}
+
+	if len(configImage.Status.ExternalRegistryHostnames) == 0 {
+		return observedConfig, errs
+	}
+
+	if err := unstructured.SetNestedStringSlice(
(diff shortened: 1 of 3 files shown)


wzheng1:
/label qe-approved

$ oc extract secrets/default-dockercfg-dbn69 --to=- | jq 'keys'
# .dockercfg
[
  "172.30.148.19:5000",
  "default-route-openshift-image-registry.apps.ci-ln-2tc5v1t-f76d1.origin-ci-int-gce.dev.openshift.com",
  "image-registry.openshift-image-registry.svc.cluster.local:5000",
  "image-registry.openshift-image-registry.svc:5000"
]
zhengwenjings-MacBook-Pro:4.0 wzheng$ oc get routes
NAME            HOST/PORT                                                                                             PATH   SERVICES         PORT    TERMINATION   WILDCARD
default-route   default-route-openshift-image-registry.apps.ci-ln-2tc5v1t-f76d1.origin-ci-int-gce.dev.openshift.com          image-registry   <all>   reencrypt     None
myregistry      registry.ci.openshift.org                                                                                    image-registry   <all>   reencrypt     None

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -52,3 +53,53 @@ func ObserveInternalRegistryHostname(genericListers configobserver.Listers, reco
...
+	if errors.IsNotFound(err) {
+		klog.V(2).Infof("images.config.openshift.io/cluster: not found")
+		return observedConfig, errs
+	} else if err != nil {
+		return prevObservedConfig, append(errs, err)
+	}
+
+	sset := sets.NewString()
+	sset.Insert(configImage.Spec.ExternalRegistryHostnames...)
+	sset.Insert(configImage.Status.ExternalRegistryHostnames...)
```
adambkaplan, line 94: Potential things to consider when using both `spec` and `status` to populate the external urls:

1. If the cluster admin adds external hostnames to spec, openshift-controller-manager has a rollout before the external hostname is reflected in the image status. I believe this doesn't happen until the route for the external hostname is created.
2. Likewise if the cluster admin removes external hostnames, openshift-controller-manager will retain "bad" external hostnames as long as the deleted external hostname is in status.

I'm inclined to only use status to set the `registryURLs` field.

adambkaplan (commented):
Generally looks good. Added a comment about using spec and status to set the controller config value.

Technically this doesn't need to be in a separate method, but given that we are using listers to access the cluster config data I don't see any harm. Keeping the reconcile logic for the internal and external hostnames separate helps with future maintenance.

pkg/operator/configobservation/images/observe_images.go
ricardomaraschini: That is interesting. During my tests I have noticed that the Status external hostname was not being kept in sync with the Spec (was planning to open a BZ for that). Turns out that the software is way clever than I am (I haven't created the route). I will push a change to use only Status. Thanks for that.

ricardomaraschini: Done, PTAL.

pkg/operator/configobservation/images/observe_images_test.go
```diff
@@ -45,3 +46,46 @@ func TestObserveRegistryConfig(t *testing.T) {
...
+				"default-route-2.openshift-image-registry.io",
+			},
+		},
+		Status: configv1.ImageStatus{
+			ExternalRegistryHostnames: []string{
+				"default-route-0.openshift-image-registry.io",
+				"default-route-1.openshift-image-registry.io",
+			},
+		},
+	}
```
adambkaplan, line 72: Please add the following test cases:

1. No external hostname (ensure we don't populate the Registry URLs)
2. External hostnames in `Spec`, but not `Status` (ensure we don't populate the registry URLs)
3. External hostnames in `Status`, but not `Spec` (ensure we keep the registry URLs)

Table-driven testing is highly effective for this purpose, and Golang testing makes it very easy to write:

https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

adambkaplan (changes requested):

pkg/operator/configobservation/images/observe_images_test.go
ricardomaraschini: done, PTAL.

adambkaplan (approved):
/lgtm

PR #207, opened 2021-04-14
WIP: set Upgradeable==true so as to no block minor updates

gabemontero:


gabemontero:
Want to run a cluster bot cluster against this PR to compare what I saw with vanilla 4.8 per https://bugzilla.redhat.com/show_bug.cgi?id=1948011#c2

Based on results we'll go from there

PR #210, opened 2021-05-04
Bug 1948011: seed upgradeable condition in ocm-o cluster operator status so library-go union code can find it

gabemontero:
/assign @adambkaplan 

ocm-o needs to minimally place the upgradeable condition in its condition array before calling into library-go, otherwise the condition union code at https://github.com/openshift/library-go/blob/d4cedcea81bbe15f5830d07419757597466544e7/pkg/operator/status/condition.go#L32 will skip it and mark it as unknown at https://github.com/openshift/library-go/blob/d4cedcea81bbe15f5830d07419757597466544e7/pkg/operator/status/condition.go#L48

The union code override effectively neuters the code at https://github.com/openshift/library-go/blob/2df205bdd55006a63d3fd22099db73432a0f9bd8/pkg/operator/status/status_controller.go#L195

### DIFF
--- a/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
+++ b/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
@@ -161,6 +161,11 @@ func syncOpenShiftControllerManager_v311_00_to_latest(c OpenShiftControllerManag
 		})
 	}
 
+	v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorapiv1.OperatorCondition{
+		Type:   operatorapiv1.OperatorStatusTypeUpgradeable,
+		Status: operatorapiv1.ConditionTrue,
+	})
+
 	operatorConfig.Status.ObservedGeneration = operatorConfig.ObjectMeta.Generation
 	resourcemerge.SetDaemonSetGeneration(&operatorConfig.Status.Generations, actualDaemonSet)
 
--- a/test/framework/conditionhelper.go
+++ b/test/framework/conditionhelper.go
@@ -16,6 +16,7 @@ func hasExpectedClusterOperatorConditions(status *configv1.ClusterOperator) bool
 	gotAvailable := false
 	gotProgressing := false
 	gotDegraded := false
+	gotUpgradeable := false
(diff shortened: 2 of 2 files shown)


gabemontero:
hit github throttling on the bugzilla bot update ... will try again later

gabemontero:
both e2e-aws and e2e-aws-upgrade hit https://bugzilla.redhat.com/show_bug.cgi?id=1956768

that bz has merged since then, let's see how those jobs look now

/retest

gabemontero:
the sig-networking pods failed to create sandboxes with mutliple bzs on sippy hit this time

wewang58:
Launched a cluster and verified it.
/lgtm

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -161,6 +161,11 @@ func syncOpenShiftControllerManager_v311_00_to_latest(c OpenShiftControllerManag
 		})
 	}
 
+	v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorapiv1.OperatorCondition{
+		Type:   operatorapiv1.OperatorStatusTypeUpgradeable,
+		Status: operatorapiv1.ConditionTrue,
+	})
```
adambkaplan, line 167: This should not affect current operator behavior.

Per the godoc:

```go
	// Upgradeable indicates whether the operator is in a state that is safe to upgrade. When status is `False`
	// administrators should not upgrade their cluster and the message field should contain a human readable description
	// of what the administrator should do to allow the operator to successfully update.  A missing condition, True,
	// and Unknown are all treated by the CVO as allowing an upgrade.
```

adambkaplan (commented):
/approve

We should also update the e2e operator test so we verify that Upgradeable is set correctly:

See https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/d7d1d793aaf88a5d250cc29d4877ec2a829973b3/test/framework/conditionhelper.go#L15-L30

gabemontero:
@wewang58 if you have cycle please reapply your lgtm

adambkaplan (approved):
/lgtm

PR #216, opened 2021-06-01
Bug 1965283: Static Resources Controller for Sync 

adambkaplan:
- Refactor ocm-o to use the static resources controller to sync static
  YAML assets.
- Add infomers for cluster-scoped objects, openshift-infra, and
  kube-system as a namespace to use an informer to fetch core k8s
  objects.
- Use cached ConfigMapsGetter in the main workload sync controller
- Don't set unmanaged/removed messages. The status controller is now
  responsible for setting the cluster operator states for managed/
  unmanaged. The existing code now collides with the new static
  resource sync controller.
- Increase rate limit for k8s API calls. This ensures the static
  resource sync controller is not throttled when reconciling RBAC
  objects.


pkg/operator/operator.go
```diff
@@ -51,7 +50,7 @@ func NewOpenShiftControllerManagerOperator(
 	targetImagePullSpec string,
 	operatorConfigInformer operatorinformersv1.OpenShiftControllerManagerInformer,
 	proxyInformer configinformerv1.ProxyInformer,
-	kubeInformersForOpenshiftControllerManager informers.SharedInformerFactory,
+	kubeInformers v1helpers.KubeInformersForNamespaces,
```
adambkaplan, line 53: We use this struct with the static resource sync controller and the resource sync controller, too.

pkg/operator/operator.go
```diff
@@ -89,35 +92,6 @@ func (c OpenShiftControllerManagerOperator) sync() error {
...
-		// do not change the current status as part of noting that unmanaged/removed are not supported
-		condition.Reason = reasonString
-		condition.Message = messageString
-		v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, condition)
-	}
-	if !equality.Semantic.DeepEqual(operatorConfig.Status, originalOperatorConfig.Status) {
-		if _, err := c.operatorConfigClient.OpenShiftControllerManagers().UpdateStatus(context.TODO(), operatorConfig, metav1.UpdateOptions{}); err != nil {
-			return err
-		}
-	}
```
adambkaplan, line 120: The status sync controller takes care of the CVO status reporting for us if the operator moves to Unamanged or Removed.

pkg/operator/starter.go
```diff
@@ -16,20 +17,33 @@ import (
...
+	// Increase QPS and burst to avoid client-side rate limits when reconciling RBAC.
+	// TODO: improve use of listers in the static resource sync controller
+	highRateLimitProtoKubeConfig := rest.CopyConfig(controllerConfig.ProtoKubeConfig)
+	if highRateLimitProtoKubeConfig.QPS < 50 {
+		highRateLimitProtoKubeConfig.QPS = 50
+	}
+	if highRateLimitProtoKubeConfig.Burst < 100 {
+		highRateLimitProtoKubeConfig.Burst = 100
+	}
+	kubeClient, err := kubernetes.NewForConfig(highRateLimitProtoKubeConfig)
```
adambkaplan, line 42: The static resource sync controller unfortunately doesn't use listers when reconciling RBAC objects.
This is problematic for ocm because we reconcile a lot of roles/clusterroles and their role bindings.
Long term we should be the ones to improve this in library-go, but for the interim we can increase the QPS and reduce the instances where the sync controller is throttled.

pkg/operator/starter.go
```diff
@@ -39,15 +53,22 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
 	}
 
-	kubeInformers := v1helpers.NewKubeInformersForNamespaces(kubeClient, util.TargetNamespace, util.OperatorNamespace, util.UserSpecifiedGlobalConfigNamespace)
+	kubeInformers := v1helpers.NewKubeInformersForNamespaces(kubeClient,
+		"",
+		util.TargetNamespace,
+		util.OperatorNamespace,
+		util.UserSpecifiedGlobalConfigNamespace,
+		util.InfraNamespace,
+		metav1.NamespaceSystem,
```
adambkaplan, line 62: New additions to this list:

- Cluster scoped (`""`)
- `openshift-infra`
- `kube-system`

pkg/operator/starter.go
```diff
@@ -105,10 +126,40 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		controllerConfig.EventRecorder,
 	)
 
+	staticResourceController := staticresourcecontroller.NewStaticResourceController(
```
adambkaplan, line 129: This is a simple controller that takes a bag of YAMLs and applies them, and reports status to the operator config object's status. That gets rolled up to the CVO object via the status sync controller.

adambkaplan (commented):
/assign @gabemontero 

I believe you had last touched the bits around the status reporting - this was changed a bit with the addition of the new controller.

pkg/operator/operator.go
gabemontero: @adambkaplan - would it be possible to specify the repos(s) / packages(s) where the static resource sync controller and resource sync controller are located?  I'm guessing they are in library-go, but some quick scans are not proving fruitful to me.

I'd like to do some cross referencing as part of the review.

Thanks

pkg/operator/starter.go
```diff
@@ -39,15 +53,22 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		return err
 	}
 
-	kubeInformers := v1helpers.NewKubeInformersForNamespaces(kubeClient, util.TargetNamespace, util.OperatorNamespace, util.UserSpecifiedGlobalConfigNamespace)
+	kubeInformers := v1helpers.NewKubeInformersForNamespaces(kubeClient,
+		"",
```
gabemontero, line 57: let's put a comment in code to indicate this means cluster scoped, as you clarified in the PR comment https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/216/commits/09eb355f0d0a7866c17b28f5a340936a7326eeb5#r644180187

pkg/operator/operator.go
gabemontero: wait I may have found it later in the review ... https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/216/commits/09eb355f0d0a7866c17b28f5a340936a7326eeb5#diff-0d623dfd885adb20f991bda4c2453aebd732ca6dbb4d1d4be6e79805c3b48de6R129

pkg/operator/starter.go
gabemontero: So you are specifically referring to https://github.com/openshift/library-go/blob/c537a5b2af0196a9a235ccecc0d05c2583899f3d/pkg/operator/resource/resourceapply/generic.go#L150 as an example @adambkaplan, correct ?

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -38,43 +37,11 @@ func syncOpenShiftControllerManager_v311_00_to_latest(c OpenShiftControllerManag
...
-		if currResult.Error != nil {
-			errors = append(errors, fmt.Errorf("%q (%T): %v", currResult.File, currResult.Type, currResult.Error))
-			continue
-		}
 
-		if currResult.Changed && resourcesThatForceRedeployment.Has(currResult.File) {
-			forceRollout = true
-		}
-	}
+	// TODO - use labels/annotations to force a daemonset rollout
```
gabemontero, line 41: IIRC this is a different TODO comment from what you had in your last PR in the same spot, and this is a TODO comment you want to keep, correct @adambkaplan ?

pkg/util/consts.go
```diff
@@ -6,6 +6,7 @@ const (
 	MachineSpecifiedGlobalConfigNamespace = "openshift-config-managed"
 	TargetNamespace                       = "openshift-controller-manager"
 	OperatorNamespace                     = "openshift-controller-manager-operator"
+	InfraNamespace                        = "openshift-infra"
```
gabemontero, line 9: As an alternative to my comment at https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/216/commits/09eb355f0d0a7866c17b28f5a340936a7326eeb5#diff-0d623dfd885adb20f991bda4c2453aebd732ca6dbb4d1d4be6e79805c3b48de6R57 we could add a constant here named something like `ClusterScoped` that is set to the emtpy string ""

pkg/operator/starter.go
gabemontero: If so, a commit specific link to the file in question might help us to quickly remember where to go if we ever get to this TODO

gabemontero (commented):
Nice pulling in a relatively newer library-go function @adambkaplan 

Minor comment / constant suggestions around the parameter for cluster scoped and the future TODO wrt rbac listers in the static resource controller.  Otherwise just a clarifying question or two.

And per you comment that I was last in this code, I certainly did not see anything that disrupted my recent `Upgradeable` change.

pkg/operator/starter.go
gabemontero: Or go with https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/216#discussion_r644215148 instead

pkg/operator/operator.go
adambkaplan: yep, that is it

pkg/operator/starter.go
adambkaplan: I think if I update the comment to use `StaticResourceController` then we should be ok (it is initialized later in this method).

pkg/operator/starter.go
adambkaplan: I'll add a comment to say "empty means cluster-scoped"

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
adambkaplan: Yes - we are passing `forceRollout` to a method that has been deprecated. We should use labels or annotations to drive rollouts.

pkg/operator/starter.go
gabemontero: sorry I'm not following how updating the comment to use `StaticResourceController` pertains to my question around confirming where the shortcoming in library-go is and ask to somehow point to the location in that code for future reference

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
gabemontero: thanks for the confirmation

pkg/operator/starter.go
gabemontero: sounds good thanks

pkg/operator/starter.go
adambkaplan: Updated the comments - hopefully things are a little bit clearer.

pkg/operator/starter.go
adambkaplan: updated

pkg/operator/starter.go
adambkaplan: added the TODO here to make it clear where in library-go we need to add a future enhancement.

adambkaplan (commented):
Added additional doc comments to clarify what I added (and why)

PR #221, opened 2021-06-28
Fix config merge order

### LINKED ISSUE #218: unsupportedConfigOverrides does not override config
### Expectation

Config override happens in this order: defaults < observedConfig < unsupportedConfigOverrides. With later override former. 

```
'unsupportedConfigOverrides holds a sparse config that will override any previously set options.  It only needs to be the fields to override it will end up overlaying in the following order: 1. hardcoded defaults 2. observedConfig 3. unsupportedConfigOverrides'
```
### Actual

Changes to `unsupportedConfigOverrides` doesn't override `observedConfig`. 

[This](https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/286c157755c8120ce969c7c15f969737b46b1a0a/pkg/operator/sync_openshiftcontrollermanager_v311_00.go#L186) could be reason. Shouldn't the order be as stated above?

OKD release: `4.6.0-0.okd-2021-02-14-205305`

ms-choudhary:
* unsupportedConfigOverrides should override observedConfig
Fixes #218


adambkaplan (commented):
/ok-to-test

adambkaplan (approved):
/lgtm

PR #246, opened 2022-07-21
Add namespace and RBAC needed for ingress-to-route

ravisantoshgudimetla:
Seeing errors like:

```
E0721 16:07:56.992638       1 leaderelection.go:330] error retrieving resource lock openshift-route-controller-manager/openshift-route-controllers: leases.coordination.k8s.io "openshift-route-controllers" is forbidden: User "system:serviceaccount:openshift-infra:ingress-to-route-controller" cannot get resource "leases" in API group "coordination.k8s.io" in the namespace "openshift-route-controller-manager"
```

with https://github.com/openshift/openshift-controller-manager/pull/230


bindata/v3.11.0/openshift-controller-manager/ingress-to-route-controller-clusterrole.yaml
```diff
@@ -12,6 +12,14 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
```
atiratree, line 15: hm, this will make it possible to reach leases in all namespaces since we are creating cluster role binding here https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/bindata/v3.11.0/openshift-controller-manager/ingress-to-route-controller-clusterrolebinding.yaml . Not sure this is what we want

bindata/v3.11.0/openshift-controller-manager/ingress-to-route-controller-rolebinding.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
atiratree, line 1: at the current stage this is not required since we already have the cluster role binding. But we probably should use this binding roleRef to a new basic Role with just the leases

bindata/v3.11.0/openshift-controller-manager/route-controller-ns.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-route-controller-manager
+  annotations:
+    openshift.io/node-selector: ""
+    workload.openshift.io/allowed: "management"
+  labels:
+    openshift.io/cluster-monitoring: "true"
```
atiratree, line 9: newlines

atiratree (commented):
working, but we need to tune the rbac a bit more

atiratree:
will do some RBAC testing on Monday to be sure we have it correct

atiratree:
opened a new PR that should fix the issues mentioned above: https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/247

atiratree:
#247 merged, closing this one
/close

PR #248, opened 2022-07-25
Bug 2110629: Set openshift.io/run-level to nil in openshift-controller-manager namespace

LalatenduMohanty:
The run-level was removed in https://github.com/openshift/cluster-openshift-controller-manager-operator/commit/d12a6ff349028b2d484643f8bf22e1377a2d378a
But because how the cluster-version operator reconciles manifest labels,
 dropping a label from the manifest does not remove it from the in-cluster resource
 when old clusters are updated into the new manifest.
This commit uses the approach the cluster-version operator used to drop its run-level,
 by setting the value to an empty string, which the run-level-consuming code treats identically to an unset label.

This avoids errors about:
```
...container has runAsNonRoot and image will run as root...
```
when updating to 4.11.

Signed-off-by: Lalatendu Mohanty <lmohanty@redhat.com>

### DIFF
--- a/bindata/v3.11.0/openshift-controller-manager/ns.yaml
+++ b/bindata/v3.11.0/openshift-controller-manager/ns.yaml
@@ -7,3 +7,4 @@ metadata:
     workload.openshift.io/allowed: "management"
   labels:
     openshift.io/cluster-monitoring: "true"
+    openshift.io/run-level: "" # specify no run-level turns it off on install and upgrades
--- a/pkg/operator/v311_00_assets/bindata.go
+++ b/pkg/operator/v311_00_assets/bindata.go
@@ -522,6 +522,7 @@ metadata:
     workload.openshift.io/allowed: "management"
   labels:
     openshift.io/cluster-monitoring: "true"
+    openshift.io/run-level: "" # specify no run-level turns it off on install and upgrades
 `)
 
 func v3110OpenshiftControllerManagerNsYamlBytes() ([]byte, error) {


manifests/00_namespace.yaml
```diff
@@ -9,4 +9,5 @@ metadata:
     workload.openshift.io/allowed: "management"
   labels:
     openshift.io/cluster-monitoring: "true"
+    openshift.io/run-level: "" # specify no run-level turns it off on install and upgrades
```
bparees, line 12: this is touching the wrong NS, per:
https://coreos.slack.com/archives/C03QED56GDS/p1658765567556089?thread_ts=1658523406.064569&cid=C03QED56GDS

bparees: @sdodson can you make sure we get this right (matches the changes you made to your cluster)

sdodson: Yes, this is the wrong namespace, needs to be `openshift-controller-manager` namespace. Unsure what creates and manages that.

sdodson:
@LalatenduMohanty looks like we need to update https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/bindata/v3.11.0/openshift-controller-manager/ns.yaml instead and then rebuild bindata.

manifests/00_namespace.yaml
LalatenduMohanty: yeah, I am changing the wrong namespace, will fix the PR

LalatenduMohanty: Fixed

sdodson:
@LalatenduMohanty Needs `make update-bindata` and commit that

manifests/00_namespace.yaml
wking: I'd scoped (rhbz#2110629)[https://bugzilla.redhat.com/show_bug.cgi?id=2110629] at both the operator and operand namespaces, but yes, we only need to fix the operand namespace to unblock 4.11.  Did we want to reduce scope on the bug, or are we ok fixing both namespaces in this pull?

soltysh (approved):
/lgtm
/approve

PR #249, opened 2022-07-25
Bug 2110715: Set openshift.io/run-level to nil in openshift-controller-manager namespace

LalatenduMohanty:
Backport  of #248


soltysh (approved):
/lgtm
/approve

PR #250, opened 2022-07-28
BUILD-417: Adding leader election leases

coreydaley:


bindata/v3.11.0/openshift-controller-manager/leader-role.yaml
```diff
@@ -22,3 +22,11 @@ rules:
   - create
   - update
   - patch
+- apiGroups:
+  - "coordination.k8s.io"
+  resources:
+  - leases
+  verbs:
+  - get
```
akram, line 30: can you confirm that you don't need `list` and `watch` permissions as well @coreydaley ?

apart from that, it ltgm

akram (approved):

bindata/v3.11.0/openshift-controller-manager/leader-role.yaml
coreydaley: The status quo with other pull requests adding this permission include them, so I think we should add them also. Can you add the lgtm if you find this acceptable?

PR #251, opened 2022-07-28
[release-4.11] Bug 2111901: Add namespace and RBAC needed for ingress-to-route

atiratree:
cherry-pick of #247

PR #255, opened 2022-08-10
Manage route controllers in separate ns

atiratree:
following up on https://github.com/openshift/openshift-controller-manager/pull/234 

TODO:
- [ ] #256 

4.11 backports that introduce leases
- [x] https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/251 
- [x] https://github.com/openshift/openshift-controller-manager/pull/232


pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -33,75 +33,128 @@ import (
...
-	if actualDaemonSet == nil {
+	if actualDaemonSet == nil || actualRCDeployment == nil {
 		return syncReturn(c, errors, originalOperatorConfig, operatorConfig)
 	}
+	available := actualDaemonSet.Status.NumberAvailable > 0
+	rcAvailable := actualRCDeployment.Status.AvailableReplicas > 0
 
 	// manage status
-	if actualDaemonSet.Status.NumberAvailable > 0 {
+	if available && rcAvailable {
```
atiratree, line 117: sharing the conditions so both need to be available to flag available as true

bindata/v3.11.0/openshift-controller-manager/route-controller-deploy.yaml
```diff
@@ -0,0 +1,109 @@
...
+        ports:
+          - containerPort: 8443
+        terminationMessagePolicy: FallbackToLogsOnError
+        volumeMounts:
+        - mountPath: /var/run/configmaps/config
+          name: config
+        - mountPath: /var/run/configmaps/client-ca
+          name: client-ca
+        - mountPath: /var/run/secrets/serving-cert
+          name: serving-cert
```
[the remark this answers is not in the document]
deads2k: > this is missing the proxy-ca-bundles and proxy envs compared to the original DS as I have not seen/found a use case for these

acceptable.

bindata/v3.11.0/openshift-controller-manager/route-controller-deploy.yaml
```diff
@@ -0,0 +1,109 @@
...
+        livenessProbe:
+          initialDelaySeconds: 30
+          httpGet:
+            scheme: HTTPS
+            port: 8443
+            path: healthz
+        readinessProbe:
+          failureThreshold: 10
```
deads2k, line 68: we should come back to this post-merge.  it looks too long.

pkg/operator/operator.go
```diff
@@ -52,18 +67,22 @@ func NewOpenShiftControllerManagerOperator(
...
-		rateLimiter:          flowcontrol.NewTokenBucketRateLimiter(0.05 /*3 per minute*/, 4),
-		recorder:             recorder,
+		targetImagePullSpec:       targetImagePullSpec,
+		operatorConfigClient:      operatorConfigClient,
+		countNodes:                countNodes,
+		ensureAtMostOnePodPerNode: ensureAtMostOnePodPerNode,
+		proxyLister:               proxyInformer.Lister(),
+		kubeClient:                kubeClient,
+		configMapsGetter:          v1helpers.CachedConfigMapGetter(kubeClient.CoreV1(), kubeInformers),
+		queue:                     workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "KubeApiserverOperator"),
```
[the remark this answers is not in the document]
deads2k: > should I open additional PR for this?

yes

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -33,75 +33,128 @@ import (
...
 	_, serviceCAModified, err := manageOpenShiftServiceCAConfigMap_v311_00_to_latest(c.kubeClient, c.configMapsGetter, c.recorder)
 	if err != nil {
-		errors = append(errors, fmt.Errorf("%q: %v", "openshift-service-ca", err))
+		errors = append(errors, fmt.Errorf("%q %q: %v", operandName, "openshift-service-ca", err))
 	}
 
 	_, globalCAModified, err := manageOpenShiftGlobalCAConfigMap_v311_00_to_latest(c.kubeClient, c.configMapsGetter, c.recorder)
 	if err != nil {
-		errors = append(errors, fmt.Errorf("%q: %v", "openshift-global-ca", err))
+		errors = append(errors, fmt.Errorf("%q %q: %v", operandName, "openshift-global-ca", err))
```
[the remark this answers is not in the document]
deads2k: This should be refactored later.  We can discuss at the control plane arch call next week.

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -180,6 +242,16 @@ func manageOpenShiftControllerManagerClientCA_v311_00_to_latest(client coreclien
 	return caChanged, nil
 }
 
+func manageRouteControllerManagerClientCA_v311_00_to_latest(client coreclientv1.ConfigMapsGetter, recorder events.Recorder) (bool, error) {
```
deads2k, line 245: this smells weird.  I think we should cleanup in the future.

deads2k:
the overall approach seems ok.  I'll leave the detailed review to @ingvagabund 

/approve

/assign @ingvagabund

pkg/operator/operator.go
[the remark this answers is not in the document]
atiratree: done: https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/257

manifests/0000_90_openshift-controller-manager-operator_04_operand-route-controller-servicemonitor.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
ingvagabund, line 2: This role is already defined in bindata/v3.11.0/openshift-controller-manager/route-controller-servicemonitor-role.yaml in https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/255/commits/d505b981d68a79a02a0038dba60e40f6647b9a5e#diff-ddf2d7b29e1f246cc23d33358c5891d71becce69bd264c2495d5f8f607a2d3a6

manifests/0000_90_openshift-controller-manager-operator_04_operand-route-controller-servicemonitor.yaml
```diff
@@ -0,0 +1,66 @@
...
+  - services
+  - endpoints
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
```
ingvagabund, line 22: This role binding is already defined in bindata/v3.11.0/openshift-controller-manager/route-controller-servicemonitor-rolebinding.yaml in https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/255/commits/d505b981d68a79a02a0038dba60e40f6647b9a5e#diff-6683cf67175abdad3fc07c757a5fe0e2c62b0c88f642989b1873bcd4f3f224a0

pkg/operator/starter.go
```diff
@@ -153,7 +154,20 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+			"v3.11.0/openshift-controller-manager/route-controller-informer-clusterrole.yaml",
+			"v3.11.0/openshift-controller-manager/route-controller-informer-clusterrolebinding.yaml",
+			"v3.11.0/openshift-controller-manager/route-controller-leader-role.yaml",
+			"v3.11.0/openshift-controller-manager/route-controller-leader-rolebinding.yaml",
 			"v3.11.0/openshift-controller-manager/route-controller-ns.yaml",
+			"v3.11.0/openshift-controller-manager/route-controller-sa.yaml",
+			"v3.11.0/openshift-controller-manager/route-controller-separate-sa-role.yaml",
+			"v3.11.0/openshift-controller-manager/route-controller-separate-sa-rolebinding.yaml",
+			"v3.11.0/openshift-controller-manager/route-controller-servicemonitor-role.yaml",
+			"v3.11.0/openshift-controller-manager/route-controller-servicemonitor-rolebinding.yaml",
```
ingvagabund, line 170: Both service monitor role and rolebinding are defined in `manifests/0000_90_openshift-controller-manager-operator_04_operand-route-controller-servicemonitor.yaml`. Will not CVO manage both role and rolebinding as well?

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -207,6 +279,32 @@ func manageOpenShiftControllerManagerConfigMap_v311_00_to_latest(kubeClient kube
 	return resourceapply.ApplyConfigMap(context.Background(), client, recorder, requiredConfigMap)
 }
 
+// similar logic for route-controller-manager in manageOpenShiftControllerManagerConfigMap_v311_00_to_latest
+func manageRouteControllerManagerConfigMap_v311_00_to_latest(kubeClient kubernetes.Interface, client coreclientv1.ConfigMapsGetter, recorder events.Recorder, operatorConfig *operatorapiv1.OpenShiftControllerManager) (*corev1.ConfigMap, bool, error) {
+	configMap := resourceread.ReadConfigMapV1OrDie(v311_00_assets.MustAsset("v3.11.0/openshift-controller-manager/route-controller-cm.yaml"))
+	defaultConfig := v311_00_assets.MustAsset("v3.11.0/config/defaultconfig.yaml")
```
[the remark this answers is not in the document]
atiratree: We use the same approach and share same config as for OCM (even after all the changes). So user can use this and disabled/enable the controllers via the overrides if they want.

Do we want to remove this functionality and start both controllers `"openshift.io/ingress-ip"` and `"openshift.io/ingress-to-route"` hardcoded every time?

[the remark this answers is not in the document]
ingvagabund: I see now. Sharing the config file across both OCM and the route controller manager. No objection here then.

pkg/operator/starter.go
ingvagabund: Checking https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/manifests/0000_90_openshift-controller-manager-operator_03_operand-servicemonitor.yaml:
```
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: prometheus-k8s
  namespace: openshift-controller-manager
  annotations:
    include.release.openshift.io/self-managed-high-availability: "true"
    include.release.openshift.io/single-node-developer: "true"
...
```
vs. https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/bindata/v3.11.0/openshift-controller-manager/servicemonitor-role.yaml:
```
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: prometheus-k8s
  namespace: openshift-controller-manager
```

The former is missing the annotations. We need to revisit the manifests later to see whether we can remove the latter to remove duplicates.

atiratree: ack, let's revisit this later

PR #256, opened 2022-08-16
fix ingress-to-route-controller RBAC

atiratree:
found some missing RBAC requirements for  ingress-to-route-controller when implementing https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/255. Till now they were fulfilled by using shared kube informers

### DIFF
--- a/bindata/v3.11.0/openshift-controller-manager/ingress-to-route-controller-clusterrole.yaml
+++ b/bindata/v3.11.0/openshift-controller-manager/ingress-to-route-controller-clusterrole.yaml
@@ -15,7 +15,8 @@ rules:
 - apiGroups:
   - networking.k8s.io
   resources:
-  - ingress
+  - ingresses
+  - ingressclasses
   verbs:
   - get
   - list
@@ -47,6 +48,7 @@ rules:
   - update
 - apiGroups:
   - ""
+  - events.k8s.io
   resources:
   - events
   verbs:
--- a/pkg/operator/v311_00_assets/bindata.go
+++ b/pkg/operator/v311_00_assets/bindata.go
@@ -362,7 +362,8 @@ rules:
 - apiGroups:
   - networking.k8s.io
   resources:
-  - ingress
+  - ingresses
+  - ingressclasses
   verbs:
   - get
   - list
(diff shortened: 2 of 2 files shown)


bindata/v3.11.0/openshift-controller-manager/ingress-to-route-controller-clusterrole.yaml
```diff
@@ -15,7 +15,8 @@ rules:
 - apiGroups:
   - networking.k8s.io
   resources:
-  - ingress
+  - ingresses
+  - ingressclasses
```
deads2k, line 19: This looks off to me.  Can you create another cluusterrole that is used to establish watches and grant write event permissions?

The permissions for this particular controller should only be those permissions needed for the control loop, which are usually the write permissions.

test/framework/clientset.go
```diff
@@ -67,7 +67,7 @@ func MustNewClientset(t *testing.T, kubeconfig *restclient.Config) *Clientset {
 // getConfig creates a *rest.Config for talking to a Kubernetes apiserver.
 // Otherwise will assume running in cluster and use the cluster provided kubeconfig.
 //
-// Config precedence
+// # Config precedence
```
deads2k, line 70: why this change?

atiratree: I get these changes when I run `make update` but CI doesn't agree with me, reverted...

bindata/v3.11.0/openshift-controller-manager/ingress-to-route-controller-clusterrole.yaml
atiratree: we are already giving these persmissions via shared informers and `openshift-controller-manager-sa` https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/bindata/v3.11.0/openshift-controller-manager/informer-clusterrole.yaml

I am doing something similar in the split https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/d1ad600a0c86a3b808e529bb622a7b2a3e33b17e/bindata/v3.11.0/openshift-controller-manager/route-controller-informer-clusterrole.yaml

this change here was only for declaring the RBAC dependency of the `ingress-to-route-controller`

PR #258, opened 2022-08-29
switch to route-controller-manager image and use ApplyDeployment

atiratree:
following up on https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/255

- use  route-controller-manager image
- remove deprecated  ApplyDeploymentWithForce and ApplyDaemonSetWithForce 

TODO:

- [x] https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/255

### DIFF
--- a/manifests/09_deployment.yaml
+++ b/manifests/09_deployment.yaml
@@ -58,6 +58,8 @@ spec:
           value: "0.0.1-snapshot"
         - name: IMAGE
           value: quay.io/openshift/origin-openshift-controller-manager:v4.2
+        - name: ROUTE_CONTROLLER_MANAGER_IMAGE
+          value: quay.io/openshift/origin-openshift-route-controller-manager:v4.12
         - name: POD_NAME
           valueFrom:
             fieldRef:
--- a/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
+++ b/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
@@ -99,7 +99,7 @@ func syncOpenShiftControllerManager_v311_00_to_latest(
 		errors = append(errors, fmt.Errorf(msg))
 	}
 
-	actualRCDeployment, _, routerControllerManagerError := manageRouteControllerManagerDeployment_v311_00_to_latest(c.kubeClient.AppsV1(), countNodes, ensureAtMostOnePodPerNodeFn, c.recorder, operatorConfig, c.targetImagePullSpec, operatorConfig.Status.Generations, rcForceRollout)
+	actualRCDeployment, _, routerControllerManagerError := manageRouteControllerManagerDeployment_v311_00_to_latest(c.kubeClient.AppsV1(), countNodes, ensureAtMostOnePodPerNodeFn, c.recorder, operatorConfig, c.routeControllerManagerTargetImagePullSpec, operatorConfig.Status.Generations, rcForceRollout)
 	if routerControllerManagerError != nil {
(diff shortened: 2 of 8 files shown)


pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -429,16 +447,11 @@ func manageOpenShiftControllerManagerDeployment_v311_00_to_latest(client appscli
 				}
 				newEnvs = append(newEnvs, env)
 			}
-			// reflect.DeepEqual does not consider this case equal
```
deads2k, line 422: why is this no longer needed?

deads2k:
the change looks reasonable to me, modulo the question. I'll leave lgtm with @ingvagabund or delegate

/approve

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
atiratree: because the new approach is to do a hash of the whole spec: https://github.com/openshift/library-go/blob/245d1ca16d1514313f3a7ab13ff2c9e3b07b5f43/pkg/operator/resource/resourceapply/apps.go#L95

and compare that here https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/apps.go#L131

manifests/09_deployment.yaml
```diff
@@ -58,6 +58,8 @@ spec:
           value: "0.0.1-snapshot"
         - name: IMAGE
```
ingvagabund, line 59: Worth renaming IMAGE into `OPENSHIFT_CONTROLLER_MANAGER_IMAGE`

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
ingvagabund: @knelasevero for awareness about this change. We need to do the same for https://github.com/openshift/cluster-kube-descheduler-operator/blob/master/pkg/operator/target_config_reconciler.go#L591 and https://github.com/openshift/secondary-scheduler-operator/blob/master/pkg/operator/target_config_reconciler.go#L276-L277. Would you please take a look?

ingvagabund (approved):

manifests/09_deployment.yaml
atiratree: I left the name for the following reasons
- it is the main part/image of all the logic managed by the operator
- people might be using this variable in their own scripts, at least I do in mine :) 
- we mention this in our dev docs https://github.com/openshift/enhancements/blob/master/dev-guide/operators.md#how-can-i-test-changes-to-an-openshift-operatoroperandrelease-component

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
knelasevero: Will do

atiratree:
[ci](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-openshift-controller-manager-operator/258/pull-ci-openshift-cluster-openshift-controller-manager-operator-master-e2e-aws-ovn/1567926889996095488) was failing, it seems SyncConfigMap can return nil in some cases - https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/258/commits/5e87d418fa9e170ee615109b9875b92427d7cd16 should fix it

atiratree:
we have the first build
/hold cancel
/retest

atiratree:
/label docs-approved
/label qe-approved
/label px-approved

failures do not seem to be connected to this change
/retest-required

PR #259, opened 2022-08-31
OCPBUGS-722: Fix NPD when accessing rote controller spec.deployment.replicas

alvaroaleman:
The field is dereferenced in the code when comparing spec against status
which currently can lead to a NPD. Update the yaml manifest with the
default the APIServer would set to avoid that.

Ref https://issues.redhat.com/browse/OCPBUGS-765
Ref https://issues.redhat.com/browse/OCPBUGS-722

/assign @atiratree @deads2k 

Sample of CI failure where this paniced: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_kubernetes/1358/pull-ci-openshift-kubernetes-master-e2e-aws-serial/1564732125813411840/artifacts/e2e-aws-serial/gather-extra/artifacts/pods/openshift-controller-manager-operator_openshift-controller-manager-operator-84b9d797f9-ldfl7_openshift-controller-manager-operator_previous.log

### DIFF
--- a/pkg/operator/v311_00_assets/bindata.go
+++ b/pkg/operator/v311_00_assets/bindata.go
@@ -782,6 +782,7 @@ metadata:
     route-controller-manager: "true"
 spec:
   # The number of replicas will be set in code to the number of master nodes.
+  replicas: 1
   strategy:
     type: RollingUpdate
     rollingUpdate:
--- a/bindata/v3.11.0/openshift-controller-manager/route-controller-deploy.yaml
+++ b/bindata/v3.11.0/openshift-controller-manager/route-controller-deploy.yaml
@@ -8,6 +8,7 @@ metadata:
     route-controller-manager: "true"
 spec:
   # The number of replicas will be set in code to the number of master nodes.
+  replicas: 1
   strategy:
     type: RollingUpdate
     rollingUpdate:


pkg/operator/v311_00_assets/bindata.go
```diff
@@ -782,6 +782,7 @@ metadata:
     route-controller-manager: "true"
 spec:
   # The number of replicas will be set in code to the number of master nodes.
+  replicas: 1
```
atiratree, line 785: @alvaroaleman thanks for the fix!

I think semantically it semantically it would be closer to set this to 0 since we need to set it later.

also might be worth it to do the nil check in the place where this was failing: https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/73dce4b8e92b992e75a45b23a08c1013211454f6/pkg/operator/sync_openshiftcontrollermanager_v311_00.go#L173

soltysh (approved):
/approve

atiratree:
let's merge now to unblock and do followups after
/lgtm

atiratree:
https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/261 focus on the NPE check in the operator

pkg/operator/v311_00_assets/bindata.go
alvaroaleman: The thinking here was that:
* By setting it in the manifest we can be sure that there won't be any NPDs now or later, even if there is or will be additional code that acceses this
* Setting it to zero might cause a change in behavior if there is any codepath that does not set the replicas

deads2k: > The thinking here was that:
> 
> * By setting it in the manifest we can be sure that there won't be any NPDs now or later, even if there is or will be additional code that acceses this
> * Setting it to zero might cause a change in behavior if there is any codepath that does not set the replicas

This is incorrect.  In the linked code, the value of `actualRCDeployment` is from client-go when an error happened, so you get a zero-value object. It doesn't have to do with this manifest's nil or not-nil value here.

PR #262, opened 2022-09-01
OCPBUGS-722: handle errors during operand creation and update

deads2k:
This fixes an NPE.  When a client call fails, the zero value of the object is returned, not a nil.  The helper method sometimes returns nil, but can also return an error an zero-value.  This means future checks using a failed create/update write invalid status into the operator status.


atiratree:
/retitle OCPBUGS-722: handle errors during operand creation and update

PR #264, opened 2022-09-21
manage openshift-controller-manager as a Deployment

atiratree:
issue: https://issues.redhat.com/browse/BUILD-495

has impact on https://github.com/openshift/openshift-controller-manager/pull/239

TODO:
- [ ] https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/258


bindata/v3.11.0/openshift-controller-manager/deploy.yaml
```diff
@@ -0,0 +1,128 @@
...
+        command: ["openshift-controller-manager", "start"]
+        args:
+        - "--config=/var/run/configmaps/config/config.yaml"
+        resources:
+          requests:
+            memory: 100Mi
+            cpu: 100m
+        ports:
+        - containerPort: 8443
+        terminationMessagePolicy: FallbackToLogsOnError
```
atiratree, line 64: I wonder if it would be anyhow useful to have higher `terminationGracePeriodSeconds` than the default `30`. In normal circumstances we probably do not need that

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -43,63 +43,100 @@ func syncOpenShiftControllerManager_v311_00_to_latest(
...
 
-	forceRollout = forceRollout || operatorConfig.ObjectMeta.Generation != operatorConfig.Status.ObservedGeneration
-	forceRollout = forceRollout || configMapModified || clientCAModified || serviceCAModified || globalCAModified
-
-	rcForceRollout = rcForceRollout || operatorConfig.ObjectMeta.Generation != operatorConfig.Status.ObservedGeneration
-	rcForceRollout = rcForceRollout || rcConfigMapModified || rcClientCAModified
-
-	// our configmaps and secrets are in order, now it is time to create the DS
+	// our configmaps and secrets are in order, now it is time to create the Deployment
 	// TODO check basic preconditions here
```
atiratree, line 111: @deads2k Do you remember the story behind this this TODO and if it is still relevant?

deads2k: I don't remember

deads2k:
this looks like a good start to me.

/approve
/lgtm
/hold

hold in case @ingvagabund wants a look

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
atiratree: removed the TODO then, I hope that is okay

pkg/operator/starter.go
```diff
@@ -228,3 +234,59 @@ func (v *versionGetter) VersionChangedChannel() <-chan struct{} {
...
+	dsClient := kubeClient.AppsV1().DaemonSets(util.TargetNamespace)
+	deployClient := kubeClient.AppsV1().Deployments(util.TargetNamespace)
+
+	go wait.UntilWithContext(ctx, func(_ context.Context) {
+		// This function isn't expected to take long enough to suggest
+		// checking that the context is done. The wait method will do that
+		// checking.
+
+		// Check whether the legacy daemonset exists and is not marked for deletion
+		ds, err := dsClient.Get(ctx, resourceName, metav1.GetOptions{})
```
ingvagabund, line 255: This will generate on Get request every minute. Any way to completely exit the loop when the DS it not found?

atiratree: +1, exiting the loop now - the DS is only managed by the operator and should not be recreated again

PR #265, opened 2022-09-30
Revert "switch to route-controller-manager image and use ApplyDeployment"

stbenjam:
Reverts openshift/cluster-openshift-controller-manager-operator#258; tracked by OCPBUGS-1929

Per TRT policy, we are reverting this breaking change to get ci and/or nightly payloads flowing again.  This PR broke nightly payloads as the image was unavailable:

```
operator "cluster-openshift-controller-manager-operator" contained an invalid image-references file: no input image tag named "openshift-route-controller-manager"
```

Please follow https://docs.ci.openshift.org/docs/how-tos/onboarding-a-new-component/#product-builds-and-becoming-part-of-an-openshift-release-image and work with the ART team to get this image built and available.

CC:  @atiratree, @ingvagabund,  @deads2k

PR #267, opened 2022-10-04
switch to route-controller-manager image and use ApplyDeployment no. 2

atiratree:
#258 had to be reverted by #265 because wrong image reference was used. image name was just updated in  https://github.com/openshift/release/pull/32772. This PR fixes the image reference to be the same as the art & release one

### DIFF
--- a/manifests/image-references
+++ b/manifests/image-references
@@ -18,3 +18,7 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-deployer:v4.0
+  - name: route-controller-manager
+    from:
+      kind: DockerImage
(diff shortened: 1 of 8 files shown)


manifests/image-references
```diff
@@ -18,3 +18,7 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-deployer:v4.0
+  - name: route-controller-manager
```
ingvagabund, line 21: https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/258/files#diff-c5de8e303795d333d4d685de85438af5df83ab9e18225ecbc79b4be35281ded3R21

PR #270, opened 2022-11-14
let deployer pods patch/apply replication controllers

deads2k:
found while building https://github.com/openshift/oc/pull/1288 to debug https://github.com/openshift/kubernetes/pull/1393

This moves the clusterrole and binding to the operator.  I'll make another to remove it from the openshift-apiserver.

As a bit of maintenance work, moving all of these is a good idea.

/assign @soltysh


deads2k:
and debt reduction in https://github.com/openshift/openshift-apiserver/pull/331

soltysh (approved):
/lgtm
/approve

PR #281, opened 2023-03-20
OCPBUGS-7785: migrate to using lease objects for leader election

tkashem:


wangke19:
@coreydaley Please add label cherry-pick-approved, I cannot. Thanks.

PR #288, opened 2023-05-08
OCPBUGS-13895: [WRKLDS-730] route-controller-manager deployment updates

atiratree:
The following changes are required for https://github.com/openshift/route-controller-manager/pull/22 refactoring.

- add POD_NAME to route-controller-manager deployment
- introduce route-controller-defaultconfig and customize lease name openshift-route-controllers to override the default one supplied by library-go
- add RBAC for infrastructures which is used by library-go for configuring leader election


pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -312,8 +312,9 @@ func manageOpenShiftControllerManagerConfigMap_v311_00_to_latest(kubeClient kube
 // similar logic for route-controller-manager in manageOpenShiftControllerManagerConfigMap_v311_00_to_latest
 func manageRouteControllerManagerConfigMap_v311_00_to_latest(kubeClient kubernetes.Interface, client coreclientv1.ConfigMapsGetter, recorder events.Recorder, operatorConfig *operatorapiv1.OpenShiftControllerManager) (*corev1.ConfigMap, bool, error) {
 	configMap := resourceread.ReadConfigMapV1OrDie(v311_00_assets.MustAsset("v3.11.0/openshift-controller-manager/route-controller-cm.yaml"))
-	defaultConfig := v311_00_assets.MustAsset("v3.11.0/config/defaultconfig.yaml")
-	requiredConfigMap, _, err := resourcemerge.MergeConfigMap(configMap, "config.yaml", nil, defaultConfig, operatorConfig.Spec.ObservedConfig.Raw, operatorConfig.Spec.UnsupportedConfigOverrides.Raw)
+	ocmDefaultConfig := v311_00_assets.MustAsset("v3.11.0/config/defaultconfig.yaml")
+	rcmDefaultConfig := v311_00_assets.MustAsset("v3.11.0/config/route-controller-defaultconfig.yaml")
+	requiredConfigMap, _, err := resourcemerge.MergeConfigMap(configMap, "config.yaml", nil, ocmDefaultConfig, rcmDefaultConfig, operatorConfig.Spec.ObservedConfig.Raw, operatorConfig.Spec.UnsupportedConfigOverrides.Raw)
```
ingvagabund, line 317: How is the resulting `OpenShiftControllerManagerConfig` config gonna look like? If I read the code correctly you combines content of both `defaultconfig.yaml` and `route-controller-defaultconfig.yaml` into a single `OpenShiftControllerManagerConfig` object which is a union of both?

ingvagabund:
@atiratree if my interpretation of the code is correct, feel free to unhold

/hold
/approve
/lgtm

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
atiratree: The reason is to have default for RCM  that override the behaviour of OCM. But the RCM config should not influence OCM. Practically OCM is empty now, but if somebody introduces some configuration, it might be imporant for RCM as well. This very theoretic though, If you prefer I could split these two and have just dedicated configs for each component.

The main reason is to supply lease name to RCM since we introduced it before and we need to override the default one in library-go.

Other option is to invest in the lease migration, but that would be quite painful.

atiratree: I agree, it is better to split it and define new behaviour twice instead of a silent overtaking of values that might be surprising. 

I think it is better to express the difference via the filename than a comment so I renamed the `defaultconfig` to `openshift-controller-manager-defaultconfig` so both OCM and RCM config are on the same level.

bindata/v3.11.0/config/route-controller-defaultconfig.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: openshiftcontrolplane.config.openshift.io/v1
```
ingvagabund, line 1: route-controller-defaultconfig.yaml -> route-controller-manager-defaultconfig.yaml

atiratree: I have changed all route-controller-manager files to have a consistent prefix

PR #292, opened 2023-05-21
OCPBUGS-13926: add loglevel controller for  OCM-o

lance5890:
add loglevel controller for openshift-controller-manager-operator


lance5890:
@atiratree @ingvagabund

atiratree (approved):

atiratree:
looks good, thanks for the PR, but it should be tracked as a bugs since we should have been supporting this. But probably no backports are required since nobody complained so far as we have the v4 set.

PR #298, opened 2023-08-30
API-1642: Do not generate image pull secrets for internal registry when internal registry is disabled.

sanchezl:
If internal image registry it removed, do not create service account image pull secrets.


pkg/operator/configobservation/controllers/observe_controllers.go
```diff
@@ -0,0 +1,102 @@
...
+	controllers := append([]string{}, defaultValue...)
+	unstructured.SetNestedStringSlice(observedConfig, controllers, "controllers")
+
+	imageRegistryEnabled, err := imageRegistryIsEnabled(listers.ImageRegistryConfigLister)
+	if err != nil {
+		return previousConfig, append(errs, err)
+	}
+
+	if imageRegistryEnabled {
+		controllers = enableController(controllers, "openshift.io/serviceaccount-pull-secrets")
```
deads2k, line 54: this is the only one?  I'm surprised.

Without an imageregistry, why have  https://github.com/openshift/openshift-controller-manager/blob/master/pkg/cmd/controller/config.go#L21C26-L23 ?

Can we trigger based on the capabilities API instead?  https://github.com/openshift/api/blame/8891815aa476232109dccf6c11b8611d209445d9/config/v1/types_cluster_version.go#L252

flavianmissi: > Without an imageregistry, why have https://github.com/openshift/openshift-controller-manager/blob/master/pkg/cmd/controller/config.go#L21C26-L23 ?
These controllers are for image streams, and image streams can function without the image registry.
I'm actually not 100% sure we can remove the SA pull-secrets controller, I'll do some digging and write back when I know more.

flavianmissi: One important thing to consider is that if the registry operator is enabled later, we need to make sure that this controller will create pull secrets for existing service accounts.
cc @xiuwang @wewang58 - think we might need your help testing this, I don't think we have great automated coverage at the moment.

sanchezl: > One important thing to consider is that if the registry operator is enabled later, we need to make sure that this controller will create pull secrets for existing service accounts.

From my own testing, I can confirm that pull secrets are generated for existing service accounts.

xiuwang: @flavianmissi 
1.  I checked the secrets when registry operator is disabled, serviceaccount have no any pull secret created. but we still could pull private external images after added related auth to openshift-config/pull-secret 
```
$oc get sa
NAME       SECRETS   AGE
builder    0         4s
default    0         4s
deployer   0         4s
 $oc get secret
No resources found in wxj namespace.
```
2. Enable registry operator again, then the project pull secrets will be created
```
oc get secret -n wxj
NAME                       TYPE                                  DATA   AGE
builder-dockercfg-d44fv    kubernetes.io/dockercfg               1      38m
builder-token-9q78w        kubernetes.io/service-account-token   4      39m
default-dockercfg-mjtd6    kubernetes.io/dockercfg               1      38m
default-token-m6rtp        kubernetes.io/service-account-token   4      39m
deployer-dockercfg-646jr   kubernetes.io/dockercfg               1      38m
deployer-token-2nnxj       kubernetes.io/service-account-token   4      39m
```
Could pull/push images from/to internal registry

flavianmissi: thank you both, sounds like we're good to go then!

sanchezl:
@flavianmissi , @deads2k 

I've opened a follow-up PR https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/300 that deletes existing generated image pull secrets when the registry is disabled for review. I would like to merge it into this PR.

pkg/operator/configobservation/controllers/observe_controllers.go
```diff
@@ -0,0 +1,102 @@
...
+	"openshift.io/build",
+	"openshift.io/build-config-change",
+	"openshift.io/deployer",
+	"openshift.io/deploymentconfig",
+	"openshift.io/image-trigger",
+	"openshift.io/image-import",
+	"openshift.io/image-signature-import",
+	"openshift.io/templateinstance",
+	"openshift.io/templateinstancefinalizer",
+	"openshift.io/unidling",
```
ingvagabund, line 30: https://github.com/openshift/api/blob/master/openshiftcontrolplane/v1/types.go#L201-L216 as constants

pkg/operator/configobservation/controllers/observe_controllers.go
flavianmissi: > Can we trigger based on the capabilities API instead? https://github.com/openshift/api/blame/8891815aa476232109dccf6c11b8611d209445d9/config/v1/types_cluster_version.go#L252

I think we could _also_ trigger on the capabilities API. Using both is a good idea, as the service account pull secret controller won't be needed on either case.

pkg/operator/configobservation/controllers/observe_controllers.go
sanchezl: Done.

flavianmissi:
failure seems unrelated
/retest

flavianmissi:
lgtm from the registry perspective

flavianmissi:
Hmm I think the test is permafailing: https://search.ci.openshift.org/?search=The+HAProxy+router+should+serve+routes+that+were+created+from+an+ingress&maxAge=48h&context=1&type=bug%2Bissue%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -1,6 +1,7 @@
 package configobservercontroller
 
 import (
+	"github.com/openshift/cluster-openshift-controller-manager-operator/pkg/operator/configobservation/controllers"
```
ardaguclu, line 4: nit: This should be at the end of import

sanchezl: Done.

sanchezl:
> Hmm I think the test is permafailing: https://search.ci.openshift.org/?search=The+HAProxy+router+should+serve+routes+that+were+created+from+an+ingress&maxAge=48h&context=1&type=bug%2Bissue%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

@flavianmissi : Eventually figured out that the route-controller-manager was sharing the config with the openshift-controller-manager. Should be fixed now.

flavianmissi:
nice work!
/lgtm

pkg/operator/configobservation/controllers/capability_imageregistry.go
```diff
@@ -0,0 +1,50 @@
...
+		}
+	}
+	controllers := []openshiftcontrolplanev1.OpenShiftControllerName{
+		openshiftcontrolplanev1.OpenShiftServiceAccountPullSecretsController,
+	}
+	if !imageRegistryCapabilityEnabled {
+		return controllers, nil
+	}
+
+	co, err := listers.ClusterOperatorLister.Get("image-registry")
```
soltysh, line 32: Why do we need those extra checks for internal image registry? We should not look into other operators information, only capabilities should drive the behavior.

soltysh: I'd also suggest squashing those 3 (or at least 2) identical methods into a single function, and a map holding information which capabilities match which controllers. This way you'll have only a single method passing a capability and in return you'll get a list of controllers to enable.

soltysh (changes requested):
I need a clear information about why we're looking into internal registry's cluster operator before merging this.

pkg/operator/configobservation/controllers/capability_imageregistry.go
sanchezl: > I'd also suggest squashing those 3 (or at least 2) identical methods into a single function, and a map holding information which capabilities match which controllers. This way you'll have only a single method passing a capability and in return you'll get a list of controllers to enable.

I used different methods for readability and maintainability as they represented three capabilities across three teams, and one of the cases was more complicated and needed more logic to compute.

soltysh (approved):
/approve

PR #291, opened 2023-05-15
WRKLDS-727: Capabilities: avoid running controllers when capabilities are not enabled

knelasevero:
Title says it all

PR #299, opened 2023-09-11
OCPBUGS-18498: Disable BuildConfigChange controller when Build cap is disabled

ardaguclu:
With this https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/291, we have an ability to disable controllers if the capability they are belonging to is disabled. 

But BuildConfigChange controller also belongs to Build capability and this PR adds it to the capability list. If Build cap
is disabled, openshift-controller-manager should not run BuildConfigChange controller.


ardaguclu:
/hold
we verify that this works;

/cc @ingvagabund

ardaguclu:
/jira refresh
I think, after the verification, we need to backport this PR and the next PR(another one for informers) in 4.14.

ardaguclu:
This is the other PR https://github.com/openshift/openshift-controller-manager/pull/270

ingvagabund:
This is a critical fix as vanilla clusters with Build/DC cap disabled will cause the OCM to restart.

PR #302, opened 2023-09-14
OCPBUGS-18932: Always sort disabled controller list

ardaguclu:
Currently, controller list is unsorted and any order change in that list (even though the values are not changed) causes new OCM roll out.

This is happening only when the capabitilies are disabled because if there is no disabled capability, controller list has only 1 item.

This commit always sorts this list for consistency and prevent unnecessary rollouts.

### DIFF
--- a/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
+++ b/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"sort"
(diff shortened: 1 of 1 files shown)


ardaguclu:
/hold
until it is verified that patch solves the problem.

ardaguclu:
/hold cancel
OCM hasn't been rolled out without any explicit reason which proves that patch is working.

ingvagabund:
/approve
/lgtm
This is a critical fix as vanilla clusters with Build/DC cap disabled will cause the OCM to restart.

PR #306, opened 2023-10-12
OCPBUGS-20164: Include Build CRD in manifests

adambkaplan:
Include the build CRD in ocm-o's manifests, so it is decoupled by the cluster-config operator. This allows the "Build" capability to be respected.


adambkaplan:
@bparees @wking removed WIP. Per our discussion, we think this needs to land before https://github.com/openshift/cluster-config-operator/pull/360

bparees:
/approve

will let @wking do the lgtm honors/double-check

wking (approved):
/lgtm

PR #315, opened 2023-11-06
OCPBUGS-22956: Remove blockage of ConfigObserver by build informer HasSynced flag

ardaguclu:
ConfigObserver is blocked until build informer is marked as synced.
However, if build capability is disabled, this blocks the start of
ConfigObserver infinitely. 

This PR starts a simple goroutine to watch capability changes and
if a capability is enabled(i.e. build), it restarts the operator pod to
start the informers and other capability related controllers.

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -11,6 +11,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -128,6 +129,24 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		controllerConfig.EventRecorder,
 	)
 
+	if !cache.WaitForCacheSync(ctx.Done(), configInformers.Config().V1().ClusterVersions().Informer().HasSynced) {
+		klog.Errorf("timed out waiting for configInformers ClusterVersions")
+		return fmt.Errorf("timed out waiting for configInformers ClusterVersions")
+	}
+
+	buildCapabilityEnabled := false
+	cv, err := configInformers.Config().V1().ClusterVersions().Lister().Get("version")
+	if err != nil {
+		return err
+	}
+
+	for _, capability := range cv.Status.Capabilities.EnabledCapabilities {
+		if capability == configv1.ClusterVersionCapabilityBuild {
+			buildCapabilityEnabled = true
+			break
+		}
+	}
+
 	// ConfigObserver observes the configuration state from cluster config objects and transforms
 	// them into configuration used by openshift-controller-manager
 	configObserver := configobservationcontroller.NewConfigObserver(
@@ -137,6 +156,7 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		kubeInformers.InformersFor(util.OperatorNamespace),
 		featureGateAccessor,
 		controllerConfig.EventRecorder,
+		buildCapabilityEnabled,
 	)
 
 	// userCAObserver watches the cluster proxy config and updates the resourceSyncer.
@@ -258,8 +278,39 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 	go logLevelController.Run(ctx, 1)
 	go imagePullSecretCleanupController.Run(ctx, 1)
 
-	<-ctx.Done()
-	return fmt.Errorf("stopped")
+	capabilityChangedCh := make(chan struct{})
+	if !buildCapabilityEnabled {
+		// check capability periodically and close chan and return
+		go func() {
+			ticker := time.NewTicker(5 * time.Minute)
+			defer ticker.Stop()
+
+			for {
+				select {
+				case <-ticker.C:
+					cv, err := configInformers.Config().V1().ClusterVersions().Lister().Get("version")
+					if err != nil {
+						klog.Errorf("capability checker error %v", err)
+						continue
+					}
+
+					for _, capability := range cv.Status.Capabilities.EnabledCapabilities {
+						if capability == configv1.ClusterVersionCapabilityBuild {
+							close(capabilityChangedCh)
+							return
+						}
+					}
+				}
+			}
+		}()
+	}
+
+	select {
+	case <-capabilityChangedCh:
+		return fmt.Errorf("capability is enabled, stopping")
+	case <-ctx.Done():
+		return fmt.Errorf("stopped")
+	}
 }
 
 type versionGetter struct {
--- a/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
+++ b/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
@@ -30,32 +30,34 @@ func NewConfigObserver(
 	kubeInformersForOperatorNamespace kubeinformers.SharedInformerFactory,
 	featureGateAccessor featuregates.FeatureGateAccess,
 	eventRecorder events.Recorder,
+	buildEnabled bool,
 ) factory.Controller {
-	c := configobserver.NewConfigObserver(
-		operatorClient,
-		eventRecorder,
-		configobservation.Listers{
-			ImageConfigLister:     configInformers.Config().V1().Images().Lister(),
-			BuildConfigLister:     configInformers.Config().V1().Builds().Lister(),
-			NetworkLister:         configInformers.Config().V1().Networks().Lister(),
-			FeatureGateLister_:    configInformers.Config().V1().FeatureGates().Lister(),
-			ClusterVersionLister:  configInformers.Config().V1().ClusterVersions().Lister(),
-			ClusterOperatorLister: configInformers.Config().V1().ClusterOperators().Lister(),
-			ConfigMapLister:       kubeInformersForOperatorNamespace.Core().V1().ConfigMaps().Lister(),
-			PreRunCachesSynced: []cache.InformerSynced{
-				configInformers.Config().V1().Builds().Informer().HasSynced,
-				configInformers.Config().V1().Images().Informer().HasSynced,
-				configInformers.Config().V1().Networks().Informer().HasSynced,
-				configInformers.Config().V1().ClusterVersions().Informer().HasSynced,
-				configInformers.Config().V1().ClusterOperators().Informer().HasSynced,
-				kubeInformersForOperatorNamespace.Core().V1().ConfigMaps().Informer().HasSynced,
-				operatorConfigInformers.Operator().V1().OpenShiftControllerManagers().Informer().HasSynced,
-			},
-		},
-		[]factory.Informer{operatorConfigInformers.Operator().V1().OpenShiftControllerManagers().Informer()},
+	informersSynced := []cache.InformerSynced{
+		configInformers.Config().V1().Images().Informer().HasSynced,
+		configInformers.Config().V1().Networks().Informer().HasSynced,
+		configInformers.Config().V1().ClusterVersions().Informer().HasSynced,
+		configInformers.Config().V1().ClusterOperators().Informer().HasSynced,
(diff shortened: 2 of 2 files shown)


bparees:
@ardaguclu is this operator restarted when the enabled capabilities on the cluster change?  if not, what will enable the informer if someone enables the build capability later?

ardaguclu:
> @ardaguclu is this operator restarted when the enabled capabilities on the cluster change? if not, what will enable the informer if someone enables the build capability later?

Thanks for review. There shouldn't be any functionality change with respect to your question ^. This PR only adds decision of adding `HasSynced` flag of build informer in informers sycned indicator or not. Please let me know, if I'm missing any apparent functionality change?

pkg/operator/starter.go
```diff
@@ -127,6 +127,19 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		controllerConfig.EventRecorder,
 	)
 
+	cv, err := configInformers.Config().V1().ClusterVersions().Lister().Get("version")
```
ingvagabund, line 130: One should wait until `configInformers.Config().V1().ClusterVersions().Informer().HasSynced` is true before accessing the object. Just to be sure to avoid NotFound error.

sanchezl (commented):
You have to be able to handle the capability being added later. I would suggest perhaps adding a `BuildConfigHasSynced cache.InformerSynced` field to `configobservation.Listers` and let the ObserveBuildControllerConfig, after determining that the BuildCapability is enabled,
 perform the check.

ardaguclu:
What would be disadvantage if this PR only removes this line https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/315/files#diff-6c6668067473c2d5cfe0007d509b3cb098b7816b9e9f2a75268798ef5d058bf3L46. It works without any problem for disabled build cap but I'd like to ensure that is there any problem for enabled build cap?

bparees:
> There shouldn't be any functionality change with respect to your question ^. This PR only adds decision of adding HasSynced flag of build informer in informers synced indicator or not. Please let me know, if I'm missing any apparent functionality change?

it's not exactly a "change in functionality", it's that to fix this bug correctly you need to do more than just not start the informer when the cap is disabled.  The code also needs to recognize when the cap is being enabled and start the informer at that point in time (or exit/kill/restart the operator process when it detects a change and let the startup logic get rerun)


If you only make the decision to start/not-start the informer when the operator starts up, then if the admin comes along later and enables the cap, the operator will continue to ignore the resource when it should be respecting it.

ardaguclu:
Build informer will run in any case, what this PR decides that ConfigObserver should be blocked until `build` informer is synced or not(via `HasSynced` flag) which is only used when the controller is started, after that all the cap changes will be managed ConfigObserver.

adambkaplan:
> Build informer will run in any case

A few questions here:

1. What is the impact if the build capability is _never_ enabled?
2. Is there any kind of exponential backoff baked into the informer running? If a cluster admin enables the `Build` capability days or weeks later, how soon with the operator respond correctly?

adambkaplan:
Also request that this bug fix be backported to OCP 4.14 - see https://issues.redhat.com/browse/OCPBUGS-20611

bparees:
> Build informer will run in any case

then this approach does not fully resolve the bug.  It might workaround the immediate issue of the container crashlooping, but a correct fix to this bug needs to ensure that this component is not trying to access optional apis when those apis are disabled (and then does start up/start accessing them if the api is enabled at a future point in time)

pkg/operator/starter.go
ardaguclu: Done.

pkg/operator/starter.go
```diff
@@ -127,6 +128,22 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+		return fmt.Errorf("unable to sync ClusterVersions informer")
+	}
+	cv, err := configInformers.Config().V1().ClusterVersions().Lister().Get("version")
+	if err != nil {
+		return err
+	}
+
+	buildEnabled := false
+	for _, capability := range cv.Status.Capabilities.EnabledCapabilities {
+		if capability == configv1.ClusterVersionCapabilityBuild {
```
wking, line 141: While cluster capabilities [cannot be disabled][1], additional cluster capabilities [can be enabled post-install][2].   So checking once at start-up time is ok, but you probably want some kind of system to trigger a container restart, or informer re-configure, or similar if `Build` is disabled when you started up but gets enabled later.  As you have it now, this controller won't notice the change until the container restarts for some other reason, and in a stable cluster that might be a long time after `Build` was enabled in ClusterVersion.

[1]: https://github.com/openshift/enhancements/blob/7af16a6363d0325d3756d2892b85bbd213b1860d/enhancements/installer/component-selection.md#capabilities-cannot-be-uninstalled
[2]: https://github.com/openshift/enhancements/blob/7af16a6363d0325d3756d2892b85bbd213b1860d/enhancements/installer/component-selection.md#capabilities-can-be-installed

ardaguclu:
After getting a couple of feedbacks about handling capability changes, I think I should be more clear;

There are 2 bugs in ConfigObserver controller;
* As a general, `ConfigObserver` can not be started (which also means that all the subcontrollers inside this also don't run) due to the `HasSynced` of build informer doesn't return anything and informer just hangs. This bug is critical because it directly affects the other components whose don't even rely on build informer. 
* BuildConfig informer is only used by `ObserveBuildControllerConfig` subcontroller and when the capability is disabled, it periodically logs not found error in here https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/5ed1e00fac146ba0b699e2d915944e022b07e095/pkg/operator/configobservation/builds/observe_builds.go#L66. `ObserveBuildControllerConfig` subcontroller should not be started when the capability is disabled and this is the second bug. But this is not high severity because it only logs not found error and doesn't cause of the crashing and can be fixed independently.

Since the first bug is critical and needs to be backported in 4.14 as soon as possible, this PR only focuses on fixing the former. The latter one requires watching the capability changes and enabling/disabling the controller, etc.(as suggested in https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/315#pullrequestreview-1715378070) and no impact on the functionality of other subcontrollers running inside `ConfigObserver`.

As you can see in the latest version of PR, this PR only removes the build informer `HasSynced` dependency. I think that even in the build enabled clusters, we still don't need `HasSynced` for build informer because worst case, it will get not found error and in second reconciliation, it will continue working.

WDYT? @ingvagabund  @bparees @adambkaplan @wking

ardaguclu:
I was wrong, container exited;

```
W1107 06:45:31.846124       1 reflector.go:535] k8s.io/client-go@v0.28.3/tools/cache/reflector.go:229: failed to list *v1.Build: the server could not find the requested resource (get builds.config.openshift.io)
E1107 06:45:31.846551       1 reflector.go:147] k8s.io/client-go@v0.28.3/tools/cache/reflector.go:229: Failed to watch *v1.Build: failed to list *v1.Build: the server could not find the requested resource (get builds.config.openshift.io)
I1107 06:45:33.632496       1 observe_builds.go:66] builds.config.openshift.io/cluster: not found
I1107 06:45:37.268551       1 observer_polling.go:111] Observed file "/var/run/secrets/serving-cert/tls.crt" has been created (hash="e3c68c12a4a04a1d25d51f84a81a08a2b213f26be26e79f6d8e1c862c12a6a73")
W1107 06:45:37.268593       1 builder.go:132] Restart triggered because of file /var/run/secrets/serving-cert/tls.crt was created
I1107 06:45:37.268640       1 observer_polling.go:111] Observed file "/var/run/secrets/serving-cert/tls.key" has been created (hash="77fbf402666ddf3e08b183d1f23982456e33d7b61bbe338c31e9396a499432e7")
I1107 06:45:37.268721       1 genericapiserver.go:681] "[graceful-termination] pre-shutdown hooks completed" name="PreShutdownHooksStopped"
I1107 06:45:37.268837       1 genericapiserver.go:538] "[graceful-termination] shutdown event" name="ShutdownInitiated"
I1107 06:45:37.269011       1 operator.go:151] Shutting down OpenShiftControllerManagerOperator
W1107 06:45:37.269051       1 builder.go:109] graceful termination failed, controllers failed with error: stopped

```

ardaguclu:
> > Build informer will run in any case
> 
> A few questions here:
> 
> 1. What is the impact if the build capability is _never_ enabled?

I'd expect that Build controller will never run

> 2. Is there any kind of exponential backoff baked into the informer running? If a cluster admin enables the `Build` capability days or weeks later, how soon with the operator respond correctly?

I'm not sure, it should fully depend on the internals of informers.

wking (approved):
I haven't confirmed the reacts-to-post-install-cap-enablement flow via Cluster Bot testing or anything, but on code inspection, your current patch looks like it will work to me :+1:

/lgtm

bparees:
> I haven't confirmed the reacts-to-post-install-cap-enablement flow via Cluster Bot testing or anything, but on code inspection, your current patch looks like it will work to me

+1

ardaguclu:
I tested it via cluster-bot no-capabilities. First I ensured that none of the informers and `BuildConfig` subcontroller in `ConfigObserver` weren't started. After that, I enabled build capability via patching `additionalcapabilities` and I saw that operator is restarted and all the informers were watching the `build resource`.

But I also saw another problem that build object is being represented as `build.build.openshift.io` instead of `build.config.openshift.io` and informers failed for a long time. I haven't dug through it but seems like a something requires a bug issue and requires investigation.

ardaguclu:
I provisioned a cluster-bot cluster with no-capabilities;
* Check that no build informer failures and build controller in ConfigObserver doesn't run
* Enable build capability via additionalcapability patching
* See that operator is restarted and informers are running and build controller is started.

and all ^ worked as expected  and this PR successfully reacts to cap enablement.

Note: Informers failed for a long time with `build.config.openshift.io` not found errors but `oc explain builds` worked (via `builds.build.openshift.io`). That might be a different another issue.

pkg/operator/starter.go
```diff
@@ -249,8 +269,38 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 	go clusterOperatorStatus.Run(ctx, 1)
 	go logLevelController.Run(ctx, 1)
 
-	<-ctx.Done()
-	return fmt.Errorf("stopped")
+	capabilityChangedCh := make(chan struct{})
+	if !buildCapabilityEnabled {
+		// check capability periodically and close chan and return
+		go func() {
+			ticker := time.NewTicker(10 * time.Minute)
```
ingvagabund, line 276: This will need a doc change stating it can take up to 10 minutes to propagate the build cap enabling to the operator.

pkg/operator/starter.go
```diff
@@ -249,8 +269,38 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+			defer ticker.Stop()
+
+			for {
+				select {
+				case <-ticker.C:
+					cv, err := configInformers.Config().V1().ClusterVersions().Lister().Get("version")
+					if err != nil {
+						klog.Errorf("capability checker error %v", err)
+					}
```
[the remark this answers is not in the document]
ardaguclu: Thanks for review @ingvagabund. Good catch and updated.

pkg/operator/starter.go
```diff
@@ -249,8 +269,39 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+				case <-ticker.C:
+					cv, err := configInformers.Config().V1().ClusterVersions().Lister().Get("version")
+					if err != nil {
+						klog.Errorf("capability checker error %v", err)
+						continue
+					}
+
+					for _, capability := range cv.Status.Capabilities.EnabledCapabilities {
+						if capability == configv1.ClusterVersionCapabilityBuild {
+							close(capabilityChangedCh)
```
ingvagabund, line 290: Would it work to cancel ctx instead of closing the additional channel?

ingvagabund:
/hold
In case others wants to review as well. `acknowledge-critical-fixes-only` needs to be set to proceed.

ardaguclu:
This is a critical fix and hold label holds the PR already;

ardaguclu:
I just re-pushed to decrease the capability checking period from 10 minutes to 5 minutes that me and @ingvagabund decided to set.

PR #320, opened 2023-11-29
OCPBUGS-24190: Disable deployer-controller when deploymentconfig is disabled

ardaguclu:
There is a deployer-controller in openshift-controller-manager that basically watches
the replicasets and pods and takes action only for the ones created by deploymentconfig.

But when deploymentconfig is disabled, this controller has no meaning. This doesn't create 
any issue because it innocently watches replicasets and pods and since there is no resource
created by deploymentconfig, it just watches forever without taking any action. 

On the other hand, there is no reason running this controller, when deploymenconfig capability
is disabled. That's why, this PR does this.


ardaguclu:
@ingvagabund could you PTAL, when you have a chance?. Thanks.

ardaguclu:
@ingvagabund kindly ping :)

PR #325, opened 2024-01-23
OCPBUGS-23624: Add .snyk file to exclude vendor and ignore unit tests

ingvagabund:
SSIA


adambkaplan:
@ingvagabund please revert/update this PR so we do not exclude the entire vendor tree.

Only vetted "false positives" should be put in the `.snyk` exclude file (with documented justification).

ardaguclu:
AFAIK Exclusion of specific files do not work on synk.

PR #346, opened 2024-05-21
OCPBUGS-34077: Always Disable Default Rolebindings Controller

adambkaplan:
With OCPBUILD-7, default namespace rolebindings for the Image Registry, DeploymentConfig, and Builds are handled by separate controllers. The initial refactor in [OCPBUILD-8](https://issues.redhat.com//browse/OCPBUILD-8) kept the default rolebindings controller in place to allow upgrade from 4.15 -> 4.16. However, the corresponding operator change in [OCPBUILD-9](https://issues.redhat.com//browse/OCPBUILD-9) forgot to disable the default rolebindings controller. This leads to the following side effects:

- On clusters with no capabilities, the default role bindings are created and reconciled.
- On a standard installation, there is potential for race conditions and server-side conflicts as two or more controllers fight to reconcile a resource. At scale, this can lead to excessive load on the apiserver.

This change ensures that the default rolebindings controller is always disabled. As a follow-up, the code for the default rolebindings controller can be safely removed in OCP 4.17.

### DIFF
--- a/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
+++ b/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
@@ -311,6 +311,14 @@ func disableControllers(clusterVersion *configv1.ClusterVersion) []string {
 	knownCaps := sets.New[configv1.ClusterVersionCapability](clusterVersion.Status.Capabilities.KnownCapabilities...)
 	capsEnabled := sets.New[configv1.ClusterVersionCapability](clusterVersion.Status.Capabilities.EnabledCapabilities...)
 
+	// OCPBUILD-8: the default pull secrets controller was refactored so each role binding can
+	// be matched with a cluster capability. The original default pull secrets controller was
+	// preserved to facilitate incremental upgrades. Starting in 4.16, the default pull secrets
+	// controller should be always be disabled to prevent race conditions and conflicts.
+
+	// TODO - the default rolebindings controller code should be removed in 4.17+
+	controllers = append(controllers, "-openshift.io/default-rolebindings")
+
 	for cont, cap := range controllerCapabilities {
 		if knownCaps.Has(cap) && !capsEnabled.Has(cap) {
 			controllers = append(controllers, fmt.Sprintf("-%s", cont))
--- a/pkg/operator/configobservation/controllers/observe_controllers.go
+++ b/pkg/operator/configobservation/controllers/observe_controllers.go
@@ -18,14 +18,18 @@ var allControllers = []string{
 	string(openshiftcontrolplanev1.OpenShiftBuildController),
 	string(openshiftcontrolplanev1.OpenShiftBuildConfigChangeController),
 	string(openshiftcontrolplanev1.OpenShiftBuilderServiceAccountController),
+	// OCPBUILD-9: the default rolebindings controller was refactored into separate controllers for
+	// the respective capability (Build, DeploymentConfig, ImageRegistry)
 	string(openshiftcontrolplanev1.OpenShiftBuilderRoleBindingsController),
 	string(openshiftcontrolplanev1.OpenShiftDeployerController),
 	string(openshiftcontrolplanev1.OpenShiftDeployerServiceAccountController),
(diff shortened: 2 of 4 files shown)


apoorvajagtap:
@adambkaplan thanks for taking care of this!
I tried testing it via a CI cluster, but the default-rolebindings controller was still not disabled. It seems [operator.disableControllers](https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go#L309) just "syncs" the controller states.

I believe, we need to append the controller details in the [controller.disabledControllers](https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/configobservation/controllers/observe_controllers.go#L63) as well.

```
func ObserveControllers(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
...
        // compile list of controllers to disable
	var disabledControllers []openshiftcontrolplanev1.OpenShiftControllerName

>>	// openshift.io/default-rolebindings controller shall be always disabled:
>>	disabledControllers = append(disabledControllers, openshiftcontrolplanev1.OpenShiftDefaultRoleBindingsController)
...
```

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -311,6 +311,14 @@ func disableControllers(clusterVersion *configv1.ClusterVersion) []string {
 	knownCaps := sets.New[configv1.ClusterVersionCapability](clusterVersion.Status.Capabilities.KnownCapabilities...)
 	capsEnabled := sets.New[configv1.ClusterVersionCapability](clusterVersion.Status.Capabilities.EnabledCapabilities...)
 
+	// OCPBUILD-8: the default pull secrets controller was refactored so each role binding can
+	// be matched with a cluster capability. The original default pull secrets controller was
+	// preserved to facilitate incremental upgrades. Starting in 4.16, the default pull secrets
+	// controller should be always be disabled to prevent race conditions and conflicts.
+
+	// TODO - the default rolebindings controller code should be removed in 4.17+
+	controllers = append(controllers, "-openshift.io/default-rolebindings")
```
adambkaplan, line 320: @ingvagabund now I am wondering why I am duplicating logic. Should this code be removed in favor of what is in `observe_controllers.go`?

adambkaplan:
/label qe-approved

@apoorvajagtap tested this on a no-capabilities cluster, then enabled Build, Image Registry, and DeploymentConfig one by one.

adambkaplan:
/label docs-approved

Documentation via release note in JIRA.

adambkaplan:
/label px-approved

No change to product experience - admin workflow to enable a capability is not changing.

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
adambkaplan: Discussed in slack - for now we are going to leave this logic in place. I may craft a follow-up PR to address this concern.

apoorvajagtap:
> /label qe-approved
> 
> @apoorvajagtap tested this on a no-capabilities cluster, then enabled Build, Image Registry, and DeploymentConfig one by one.

The `default-rolebindings` controller was disabled by default for a no-caps cluster. Adding the config cm for reference:
```
"controllers":["-openshift.io/build","-openshift.io/build-config-change","-openshift.io/builder-rolebindings",
"-openshift.io/builder-serviceaccount","-openshift.io/default-rolebindings","-openshift.io/deployer",
"-openshift.io/deployer-rolebindings","-openshift.io/deployer-serviceaccount",
"-openshift.io/deploymentconfig","openshift.io/image-import","-openshift.io/image-puller-rolebindings",
"openshift.io/image-signature-import","openshift.io/image-trigger","openshift.io/ingress-ip",
"openshift.io/ingress-to-route","openshift.io/origin-namespace","openshift.io/serviceaccount",
"-openshift.io/serviceaccount-pull-secrets","openshift.io/templateinstance",
"openshift.io/templateinstancefinalizer","openshift.io/unidling"]
```

PR #347, opened 2024-05-22
OCPBUGS-34054: lots of churn during image registry managed/removed transition

sanchezl:
Modifies the OCM-operator logic to defer the deletion of any image pull secrets that were created within the last 10 minutes. This change is required to prevent contention between the OCM v4.16 and OCM-operator, where the OCM-operator is deleting resources while the OCM is still creating them.

Also modifies the OCM-operator logic to not remove the image pull secret from the ServiceAccount.imagePullSecrets field until it is ready to delete the image pull secret. This is required to prevent contention with the OCM in v4.15 during an upgrade.

Overall, this is a stop-gap measure until we can move the image pull secret cleanup logic into the OCM itself, where we can prevent the competing control loops from running at the same time.

### DIFF
--- a/pkg/operator/internalimageregistry/cleanup_controller.go
+++ b/pkg/operator/internalimageregistry/cleanup_controller.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/openshift/client-go/config/informers/externalversions"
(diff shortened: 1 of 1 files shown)


flavianmissi (commented):
nice refactor, thanks for reacting so fast to this.
/lgtm

soltysh (approved):
/lgtm
/approve

PR #351, opened 2024-05-27
OCPBUGS-34395: Move `cluster` Build CR to runlevel 10 to match CRD

petr-muller:
CRD is applied in runlevel 10, so CR should not be applied sooner than that:

https://github.com/openshift/api/blob/ba11c1587003dc84cb014fd8db3fa597a3faaa63/config/v1/zz_generated.crd-manifests/0000_10_openshift-controller-manager_01_builds.crd.yaml


adambkaplan (approved):
/lgtm

/label qe-approved

PR #377, opened 2025-01-09
NO-JIRA: Remove marshaling in status apply

ardaguclu:
This PR removes unnecessary json marshal/unmarshal operation. Besides this PR sets condition's string fields to nil when they are passed empty string to prevent hotloop.

### DIFF
--- a/pkg/operator/generic_client.go
+++ b/pkg/operator/generic_client.go
@@ -2,16 +2,15 @@ package operator
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"github.com/openshift/library-go/pkg/apiserver/jsonpatch"
-
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/utils/clock"
+	"k8s.io/utils/ptr"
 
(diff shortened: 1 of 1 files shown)


pkg/operator/generic_client.go
```diff
@@ -141,61 +140,46 @@ func (p *genericClient) ApplyOperatorStatus(ctx context.Context, fieldManager st
 		return fmt.Errorf("applyConfiguration must have a value")
 	}
 
+	for i, val := range desiredConfiguration.Conditions {
+		// desired configuration may contain empty string fields.
+		// However, they are persisted by API server as nil. This causes a hotloop as empty string and nil
+		// are not equal. We explicitly convert empty strings to nil to prevent hotloop.
+		// This should be safe, because explicitly setting empty string has no meaning.
+		if len(ptr.Deref(val.Message, "")) == 0 {
+			desiredConfiguration.Conditions[i].Message = nil
```
ingvagabund, line 149: Nit: better to make a copy of `desiredConfiguration` before updating it to avoid side effects.

ardaguclu: It seems that `DeepCopy` isn't generated for `desiredConfiguration`.

ardaguclu:
DeepCopying would indeed be safe to avoid any side effects. However, I don't have time for now. Until when I return to this task;
/close

pkg/operator/generic_client.go
ingvagabund: As an alternative, you might find all the places where `ApplyOperatorStatus` is invoked and make sure both Message and Reason are set to `nil` before.

PR #382, opened 2025-04-25
OCPBUGS-52193: remove finalizer from legacy registry token

ricardomaraschini:
the legacy token controller manages legacy tokens. this controller runs as part of the operand for this operator, except when the registry is in a "removed" state. when in the "removed" state, the legacy token controller does not run and therefore cannot remove the finalizer.

a routine within the operator is responsible for deleting the registry secrets when in the removed state. this cleanup routine needs to also remove the finalizer before removing the secret.

PR #384, opened 2025-04-29
donotmerge: testing finalizer removal (pull secrets)

ricardomaraschini:
this pr should not be merged and is intended for testing.

### DIFF
--- a/pkg/operator/internalimageregistry/cleanup_controller.go
+++ b/pkg/operator/internalimageregistry/cleanup_controller.go
@@ -5,6 +5,7 @@ import (
 	errs "errors"
 	"fmt"
 	"net/http"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/operator/internalimageregistry/cleanup_controller.go
```diff
@@ -64,9 +64,46 @@ func (c *imagePullSecretCleanupController) sync(ctx context.Context, controllerC
...
+// finalizer if we also remove the secret pointed by its
+// openshift.io/token-secret.name annotation.
+func (c *imagePullSecretCleanupController) removeLegacyFinalizer(ctx context.Context, secret *corev1.Secret) error {
+	finalizers := []string{}
+	for _, finalizer := range secret.Finalizers {
+		if finalizer == "openshift.io/legacy-token" {
+			continue
+		}
+		finalizers = append(finalizers, finalizer)
+	}
```
sanchezl, line 92:
```suggestion
	finalizers := slices.DeleteFunc(secret.Finalizers, func(s string) bool {
		return s == "openshift.io/legacy-token"
	})
```

pkg/operator/internalimageregistry/cleanup_controller.go
```diff
@@ -64,9 +64,46 @@ func (c *imagePullSecretCleanupController) sync(ctx context.Context, controllerC
 	if imageRegistryEnabled {
 		return nil
 	}
+
+	// once the image registry is disabled we want to wait unil the controller manager
+	// operand is deployed and stable before clean up the image pull secrets. we don't
+	// want to compete with the operand that may be acting on the same resources we are.
```
sanchezl, line 70: There is still a chance for the controllers to be out of sync here, but I do think that this is an improvement.

sanchezl (changes requested):
On the right track, but this would still leave the possibility of some token secrets being missed if the  SA->ImagePullSecret->TokenSecret connections are not there. Consider bringing in the legacyImagePullSecretRollbackController into the OCM-o and running it in parallel.

pkg/operator/internalimageregistry/cleanup_controller.go
ricardomaraschini: Can you please elaborate a little bit more on the `legacyImagePullSecretRollbackController` matter ? It does not seem to be in use in the `openshift-controller-manager` project (I could not find where it is started there).

I failed to see how the controller could be out of sync, can you give me more details ?

ricardomaraschini:
This one is being superseded by https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/382

PR #385, opened 2025-05-06
WIP: WRKLDS-1676: controller manager config: Set leaderElection.name

tchap:
Set `leaderElection.name` in `OpenShiftControllerManagerConfig` to contain the custom value currently in use so that the operand deployment is backward compatible when the lease name is no longer being set in the operand executable.

This is necessary for the upcoming operand refactoring in https://github.com/openshift/openshift-controller-manager/pull/378

/hold


bindata/assets/config/openshift-controller-manager-defaultconfig.yaml
```diff
@@ -1,2 +1,4 @@
 apiVersion: openshiftcontrolplane.config.openshift.io/v1
 kind: OpenShiftControllerManagerConfig
+leaderElection:
+  name: openshift-master-controllers
```
atiratree, line 4: for posterity, this corresponds to https://github.com/openshift/openshift-controller-manager/blob/39ebafcc026c27abd1181b38e919076e3ec1b465/pkg/cmd/openshift-controller-manager/controller_manager.go#L85

atiratree:
/lgtm

@tchap btw we could also prune the old LE RBAC 

https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/0ab30996207ada63ef2e2fd4f47aac03b0872c9a/bindata/assets/openshift-controller-manager/old-leader-role.yaml#L1

tchap:
I will remove these old files. I also guess we don't need https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/0ab30996207ada63ef2e2fd4f47aac03b0872c9a/bindata/assets/openshift-controller-manager/leader-role.yaml#L14 ? I at least don't manage to find any such resource in the cluster...

bindata/assets/openshift-controller-manager/old-leader-role.yaml
```diff
@@ -1,24 +0,0 @@
-# needed to get the legacy lock that we used to use
```
atiratree, line 1: We cannot remove these files yet. We first have to delete them in existing clusters with using:

https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/0ab30996207ada63ef2e2fd4f47aac03b0872c9a/pkg/operator/starter.go#L246

The files can be removed only in the next major version of openshift at earliest.

atiratree (commented):
> I will remove these old files. I also guess we don't need

Correct, these were used during the LE migration from configmaps to leases in the past.

bindata/assets/openshift-controller-manager/old-leader-role.yaml
tchap: I moved the files into that section. Now sure how this works, I understood that you simply move the files into the other section, so that's what I did in https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/385/commits/3cc19e789dde17919d3d2cf505eb88dd811a9ded . Also rolled back the change that removed the files, naturally.

pkg/operator/starter.go
```diff
@@ -246,6 +244,8 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 	).WithConditionalResources(
 		bindata.Asset,
 		[]string{
+			"assets/openshift-controller-manager/old-leader-role.yaml",
```
atiratree, line 247: Let's also add a TODO when these should be removed

I guess we can also remove the `leader-ingress-to-route-controller` files later, once this PR merges.

atiratree: This change will be present in 4.20, so we can remove the files in 4.21.

tchap: Ammeded the last commit.

PR #386, opened 2025-05-09
WRKLDS-1676: ocm deployment: Add POD_NAME

tchap:
Also had to align deployment env generation to fix tests.

Required for https://github.com/openshift/openshift-controller-manager/pull/378

### DIFF
--- a/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
+++ b/pkg/operator/sync_openshiftcontrollermanager_v311_00.go
@@ -131,6 +131,7 @@ func syncOpenShiftControllerManager_v311_00_to_latest(
 	// our configmaps and secrets are in order, now it is time to create the Deployment
 	var progressingMessages []string
 	actualDeployment, _, openshiftControllerManagerError := manageOpenShiftControllerManagerDeployment_v311_00_to_latest(
+		bindata.MustAsset,
 		c.kubeClient.AppsV1(),
 		countNodes,
 		ensureAtMostOnePodPerNodeFn,
@@ -438,6 +439,7 @@ func manageOpenShiftGlobalCAConfigMap_v311_00_to_latest(kubeClient kubernetes.In
 }
 
 func manageOpenShiftControllerManagerDeployment_v311_00_to_latest(
+	mustLoadAsset func(string) []byte,
 	client appsclientv1.DeploymentsGetter,
 	countNodes nodeCountFunc,
 	ensureAtMostOnePodPerNodeFn ensureAtMostOnePodPerNodeFunc,
@@ -448,7 +450,7 @@ func manageOpenShiftControllerManagerDeployment_v311_00_to_latest(
 	proxyLister proxyvclient1.ProxyLister,
 	specAnnotations map[string]string,
 ) (*appsv1.Deployment, bool, error) {
-	required := resourceread.ReadDeploymentV1OrDie(bindata.MustAsset("assets/openshift-controller-manager/deploy.yaml"))
+	required := resourceread.ReadDeploymentV1OrDie(mustLoadAsset("assets/openshift-controller-manager/deploy.yaml"))
 
 	if len(imagePullSpec) > 0 {
 		required.Spec.Template.Spec.Containers[0].Image = imagePullSpec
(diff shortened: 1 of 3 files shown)


pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -493,44 +493,27 @@ func manageOpenShiftControllerManagerDeployment_v311_00_to_latest(
 		}
 	} else {
 		for i, c := range required.Spec.Template.Spec.Containers {
```
atiratree, line 495: Setting the values should be tested by https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/0ab30996207ada63ef2e2fd4f47aac03b0872c9a/pkg/operator/sync_openshiftcontrollermanager_v311_00_test.go#L780

Can we improve the test so it tests also the update part? Should be enough to just update the proxy config and call `manageOpenShiftControllerManagerDeployment_v311_00_to_latest` again.

atiratree (commented):
The code changes LGTM

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
tchap: Amended.

pkg/operator/sync_openshiftcontrollermanager_v311_00_test.go
```diff
@@ -834,58 +819,120 @@ func TestDeploymentWithProxy(t *testing.T) {
...
-	if err != nil {
-		t.Fatalf("unexpected error: %s", err.Error())
-	}
-	if !rcBool {
-		t.Fatal("apply deployment does not think a changes was made")
-	}
-
-	if ds == nil {
-		t.Fatalf("nil deployment returned")
+	tests := []struct {
```
atiratree, line 822: IMO there is no need for 2 tests

pkg/operator/sync_openshiftcontrollermanager_v311_00.go
```diff
@@ -493,44 +493,27 @@ func manageOpenShiftControllerManagerDeployment_v311_00_to_latest(
...
-				if len(proxyCfg.Status.HTTPProxy) > 0 {
-					newEnvs = append(newEnvs, corev1.EnvVar{Name: "HTTP_PROXY", Value: proxyCfg.Status.HTTPProxy})
-				}
-				if len(proxyCfg.Status.HTTPSProxy) > 0 {
-					newEnvs = append(newEnvs, corev1.EnvVar{Name: "HTTPS_PROXY", Value: proxyCfg.Status.HTTPSProxy})
+			// Remove the relevant variables from the spec.
+			newEnvs := make([]corev1.EnvVar, 0, len(c.Env))
+			for _, env := range c.Env {
+				switch strings.TrimSpace(env.Name) {
+				case "HTTPS_PROXY", "HTTP_PROXY", "NO_PROXY":
```
atiratree, line 500: basically, we just want to try to hit this branch of removing the old env variable and see that we get the new value

pkg/operator/sync_openshiftcontrollermanager_v311_00_test.go
```diff
@@ -834,58 +819,120 @@ func TestDeploymentWithProxy(t *testing.T) {
...
+			if !foundHTTPProxy {
+				t.Fatalf("HTTP_PROXY not found: %#v", ds.Spec.Template.Spec.Containers)
+			}
+			if !foundHTTPSProxy {
+				t.Fatalf("HTTPS_PROXY not found: %#v", ds.Spec.Template.Spec.Containers)
+			}
+			if !foundPodName {
+				t.Fatalf("POD_NAME not found: %#v", ds.Spec.Template.Spec.Containers)
+			}
+		})
```
atiratree, line 936: I think we can just call manageOpenShiftControllerManagerDeployment_v311_00_to_latest here again

atiratree: This is wrong - sorry for the misguidance. We do not use the cluster state when applying the deployment: https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/83746c46b6483bafad6a82832bc70b7f612a4789/vendor/github.com/openshift/library-go/pkg/operator/resource/resourceapply/apps.go#L140

pkg/operator/sync_openshiftcontrollermanager_v311_00_test.go
tchap: Testing all the cases requires multiple tests and can only be done by mocking the deployment template as it is being loaded from the asset directory. I aligned the test function to use the default setup in one case, a mocked deployment template in the second case. All cases should now be checked.

pkg/operator/sync_openshiftcontrollermanager_v311_00_test.go
```diff
@@ -778,6 +780,147 @@ func TestProgressingCondition(t *testing.T) {
...
+						if len(c.Env) != 3 {
+							t.Errorf("expected 3 env vars, found %d", len(c.Env))
+						}
+
+						switch e.Name {
+						case "NO_PROXY":
+							// This is not present in proxy config and should be removed.
+							foundNoProxy = true
+						case "HTTP_PROXY":
+							// This is set in the deployment template, but the proxy config value should be used.
```
atiratree, line 896: we can keep the comments, but move them next to the static expect fields

pkg/operator/sync_openshiftcontrollermanager_v311_00_test.go
atiratree: Ack, having multiple tests looks good now, but let's try to streamline that a bit.

tchap:
@atiratree All review issues fixed. Good points, code looks better now, will make sure to follow the same guidelines next time as well.

pkg/operator/sync_openshiftcontrollermanager_v311_00_test.go
```diff
@@ -778,114 +781,191 @@ func TestProgressingCondition(t *testing.T) {
...
+				t.Fatal("apply deployment does not think a changes was made")
+			}
+			if ds == nil {
+				t.Fatalf("nil deployment returned")
+			}
+
+			// Check the resulting env
+			for _, c := range ds.Spec.Template.Spec.Containers {
+				if !cmp.Equal(c.Env, tt.expectedEnv) {
+					t.Errorf(cmp.Diff(tt.expectedEnv, c.Env))
```
atiratree, line 966: nit: we could make the error message a little bit more verbose

atiratree (commented):
LGTM

pkg/operator/sync_openshiftcontrollermanager_v311_00_test.go
tchap: Amended.

PR #407, opened 2026-02-03
CNTRLPLANE-2620: propagate global tls config to operand

ricardomaraschini:
Observe the global apiserver tls config and propagates it to the operand in case of changes. this seems to affect only liveness and readiness probes in the operand. nonetheless getting this done will avoid a false positive later on if (tls-scanner may flag this as an issue).

This is mostly vendoring.


rikatz:
This also fixes https://issues.redhat.com/browse/OCPBUGS-62290, thanks!

PR #428, opened 2026-04-20
CNTRLPLANE-2907: e2e network policy tests

kaleemsiddiqu:
e2e network policy tests

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive NetworkPolicy e2e suites for validation, enforcement, and reconciliation (connectivity, port filtering, cross-namespace access, reconciliation); includes extensive helpers for pod readiness, connectivity probing, policy mutation/restore, diagnostics, and dynamic test registration. Test startup now fails fast on setup errors.
* **Chores**
  * Updated testing dependencies to support the enhanced e2e suites.


test/e2e/network_policy_enforcement.go
```diff
@@ -0,0 +1,236 @@
...
+	g.By("Verifying unauthorized ports are blocked on the same server (no server listening)")
+	for _, port := range []int32{80, 443, 9090, 9999, 22, 3306, 5432, 6379} {
+		g.GinkgoWriter.Printf("expecting deny from openshift-monitoring to operator:%d (unauthorized port)\n", port)
+		expectConnectivity(kubeClient, "openshift-monitoring", map[string]string{"app.kubernetes.io/name": "prometheus"}, operatorServerIPs, port, false)
+	}
+
+	g.By("Verifying unauthorized ports are blocked from default namespace")
+	for _, port := range []int32{9090, 9999} {
+		g.GinkgoWriter.Printf("expecting deny from default to operator:%d (unauthorized port)\n", port)
+		expectConnectivity(kubeClient, "default", map[string]string{"test": "any-pod"}, operatorServerIPs, port, false)
```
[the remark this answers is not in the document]
kaleemsiddiqu: This is one the negative test case to confirm that it does not works on these ports

test/e2e/helpers.go
```diff
@@ -0,0 +1,929 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// createServerPod creates a test server pod and returns its IPs and a cleanup function.
+func createServerPod(kubeClient kubernetes.Interface, namespace, name string, labels map[string]string, port int32) ([]string, func()) {
+	g.GinkgoHelper()
+	ctx := context.TODO()
```
liouk: Propagate caller's context instead of using `context.TODO()`.

liouk (changes requested):

test/e2e/helpers.go
kaleemsiddiqu: Addressed.

kaleemsiddiqu:
/verified by @kaleemsiddiqu 
local execution of these tests went fine.

kaleemsiddiqu:
/verified by @kaleemsiddiqu 
Local execution is successful

test/e2e/helpers.go
```diff
@@ -0,0 +1,966 @@
...
+		}
+		if len(rule.From) == 0 {
+			return true
+		}
+	}
+	return false
+}
+
+// namespaceSelectorMatches checks if a namespace selector matches the given namespace.
+func namespaceSelectorMatches(selector *metav1.LabelSelector, namespace string) bool {
```
ricardomaraschini: What is the difference between this function and `nsMatch` ? If there is any difference, can we consolidate into a single one ?

test/e2e/helpers.go
```diff
@@ -0,0 +1,966 @@
...
+		if labels[key] != value {
+			return false
+		}
+	}
+	return true
+}
+
+// ruleAllowsPort checks if a rule allows a specific TCP port.
+// This function is used for TCP connectivity checks, so it only matches TCP or unspecified protocol.
+func ruleAllowsPort(ports []networkingv1.NetworkPolicyPort, port int32) bool {
```
ricardomaraschini: Can we somehow make this function better integrate with `hasPort` to avoid duplication ?

test/e2e/helpers.go
```diff
@@ -0,0 +1,966 @@
...
+	})
+	if err != nil {
+		g.GinkgoWriter.Printf("connectivity %s/%s expected=%t (best-effort) failed: %v\n", namespace, formatIPPort(serverIP, port), shouldSucceed, err)
+		return
+	}
+	g.GinkgoWriter.Printf("connectivity %s/%s expected=%t (best-effort)\n", namespace, formatIPPort(serverIP, port), shouldSucceed)
+}
+
+// logConnectivityBestEffort checks connectivity to all provided IPs (best-effort).
+func logConnectivityBestEffort(ctx context.Context, kubeClient kubernetes.Interface, namespace string, clientLabels map[string]string, serverIPs []string, port int32, shouldSucceed bool) {
```
ricardomaraschini: This does not seem to be in use, according to my linter.

test/e2e/helpers.go
```diff
@@ -0,0 +1,966 @@
...
+	if line == "" {
+		return false, fmt.Errorf("no connectivity result yet from pod %s/%s", namespace, podName)
+	}
+
+	g.GinkgoWriter.Printf("client pod %s/%s result=%s\n", namespace, podName, line)
+	return line == "CONN_OK", nil
+}
+
+// ingressAllowsFromNamespace checks if a policy allows ingress from a specific namespace.
+func ingressAllowsFromNamespace(policy *networkingv1.NetworkPolicy, namespace string, labels map[string]string, port int32) bool {
```
ricardomaraschini: This does not seem to be in use, according to my linter.

test/e2e/helpers.go
```diff
@@ -0,0 +1,966 @@
...
+			if cond.Type == corev1.PodReady && cond.Status == corev1.ConditionTrue {
+				return true, nil
+			}
+		}
+		return false, nil
+	})
+}
+
+// waitForPodCompletion waits for a pod to complete (succeed or fail).
+func waitForPodCompletion(ctx context.Context, kubeClient kubernetes.Interface, namespace, name string) error {
```
ricardomaraschini: This does not seem to be in use.

test/e2e/helpers.go
```diff
@@ -0,0 +1,966 @@
...
+// requirePodSelectorLabel validates that a policy has a specific podSelector label with value.
+func requirePodSelectorLabel(policy *networkingv1.NetworkPolicy, key, value string) {
+	g.GinkgoHelper()
+	actual, ok := policy.Spec.PodSelector.MatchLabels[key]
+	o.Expect(ok && actual == value).To(o.BeTrue(),
+		"%s/%s: expected podSelector %s=%s, got %v", policy.Namespace, policy.Name, key, value, policy.Spec.PodSelector.MatchLabels)
+}
+
+// requirePodSelectorHasLabel validates that a policy has a specific podSelector label key.
+func requirePodSelectorHasLabel(policy *networkingv1.NetworkPolicy, key string) {
```
ricardomaraschini: this isn't used.

test/e2e/helpers.go
```diff
@@ -0,0 +1,966 @@
...
+		}
+	}
+	for _, expr := range selector.MatchExpressions {
+		if expr.Key != "kubernetes.io/metadata.name" {
+			continue
+		}
+		if expr.Operator != metav1.LabelSelectorOpIn {
+			continue
+		}
+		for _, value := range expr.Values {
```
ricardomaraschini: use `slice.Contains`.

test/e2e/helpers.go
```diff
@@ -0,0 +1,966 @@
...
+		if p.Protocol != nil && *p.Protocol != corev1.ProtocolTCP {
+			continue
+		}
+		return true
+	}
+	return false
+}
+
+// restoreNetworkPolicy deletes a NetworkPolicy and waits for it to be restored.
+func restoreNetworkPolicy(ctx context.Context, client kubernetes.Interface, expected *networkingv1.NetworkPolicy) {
```
ricardomaraschini: Why we have this ? Can't we go with `restoreNetworkPolicyWithTimeout()` all the time ? This might be read as `check without timeout`. I would prefer to have explicit timeouts everywhere.

test/e2e/helpers.go
```diff
@@ -0,0 +1,966 @@
...
+	namespace := expected.Namespace
+	name := expected.Name
+	originalUID := expected.UID
+	g.GinkgoWriter.Printf("deleting NetworkPolicy %s/%s (UID: %s)\n", namespace, name, originalUID)
+	err := client.NetworkingV1().NetworkPolicies(namespace).Delete(ctx, name, metav1.DeleteOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred(), "failed to delete NetworkPolicy %s/%s", namespace, name)
+
+	// Track whether deletion has been observed by checking UID change
+	deletionObserved := false
+	err = wait.PollUntilContextTimeout(ctx, 5*time.Second, timeout, true, func(ctx context.Context) (bool, error) {
```
ricardomaraschini: I have seen a lot of what seems to be "random" timeouts across this whole test. Could we adopt a unified way of aggregating these timeout into classes ? Something sorts of:

```go
const (
  defaultPollInterval = 5 * time.Second
  fastPollInterval  = 2 * time.Second
  defaultTimeout  = 2 * time.Minute
  defaultReconcileTimeout = 10 * time.Minute
  shortTimeout  = 30 * time.Second
  longTimeout  = 5 * time.Minute
)
```

I think it would make this better, what do you think ?

test/e2e/helpers.go
```diff
@@ -0,0 +1,966 @@
...
+	created, err := kubeClient.CoreV1().Pods(namespace).Get(ctx, name, metav1.GetOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred(), "failed to get created server pod")
+	o.Expect(created.Status.PodIPs).NotTo(o.BeEmpty())
+
+	ips := podIPs(created)
+	g.GinkgoWriter.Printf("server pod %s/%s ips=%v\n", namespace, name, ips)
+
+	return ips, func() {
+		g.GinkgoWriter.Printf("deleting server pod %s/%s\n", namespace, name)
+		_ = kubeClient.CoreV1().Pods(namespace).Delete(ctx, name, metav1.DeleteOptions{})
```
ricardomaraschini: Even though this is fine I would say we should log in case of errors, no ? That would make the debug easier later on. There are more errors being silently ignored that I think shouldn't. We don't need to fail necessarily but logging would be a nice to have.

test/e2e/network_policy.go
```diff
@@ -0,0 +1,141 @@
...
+	operatorPolicyName               = "allow-operator"
+)
+
+var _ = g.Describe("[Operator][NetworkPolicy] NetworkPolicy Validation", func() {
+	g.It("should have properly configured NetworkPolicies across all namespaces [Skipped:MicroShift]", func(ctx context.Context) {
+		testControllerManagerNetworkPolicies(ctx)
+	})
+})
+
+var _ = g.Describe("[Serial][Operator][NetworkPolicy] NetworkPolicy Reconciliation[Timeout:30m]", func() {
```
ricardomaraschini, line 27: How is this `Timeout` enforced ?  This function calls other functions with their own timeout that, when summed, goes way beyond the 30 minutes defined here.

test/e2e/network_policy.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+const (
+	defaultDenyAllPolicyName         = "default-deny"
+	controllerManagerPolicyName      = "allow-controller-manager"
+	routeControllerManagerPolicyName = "allow-route-controller-manager"
+	operatorPolicyName               = "allow-operator"
+)
+
+var _ = g.Describe("[Operator][NetworkPolicy] NetworkPolicy Validation", func() {
```
ricardomaraschini, line 21: Why this don't have a `Timeout` as `[Serial][Operator][NetworkPolicy] NetworkPolicy Reconciliation` ? Is the latter only "decorative" ?

test/e2e/network_policy_enforcement.go
```diff
@@ -0,0 +1,236 @@
...
+	g.It("should allow cross-namespace ingress to metrics endpoints [Skipped:MicroShift]", func(ctx context.Context) {
+		testCrossNamespaceMetricsAccess(ctx)
+	})
+	g.It("should allow metrics port 8443 but block other ports [Skipped:MicroShift]", func(ctx context.Context) {
+		testMetricsPortOpenOthersBlocked(ctx)
+	})
+})
+
+// testGenericNetworkPolicyEnforcement validates basic NetworkPolicy enforcement behavior
+// by creating a temporary namespace and testing various policy configurations.
```
ricardomaraschini: What aspect of `cluster-openshift-controller-manager-operator` are we resting here ? I understand that `testControllerManagerNetworkPolicyEnforcement` relates but this one does not seem like.

test/e2e/network_policy_enforcement.go
```diff
@@ -0,0 +1,236 @@
...
+	g.DeferCleanup(cleanupDenied)
+
+	g.By("Verifying allowed port 8443 ingress to operator")
+	expectConnectivity(ctx, kubeClient, util.OperatorNamespace, operatorLabels, allowedServerIPs, 8443, true)
+
+	g.By("Verifying denied port 12345 (not in NetworkPolicy)")
+	expectConnectivity(ctx, kubeClient, util.OperatorNamespace, operatorLabels, deniedServerIPs, 12345, false)
+
+	g.By("Verifying denied ports even from same namespace")
+	for _, port := range []int32{12346, 12347, 12348, 12349} {
```
ricardomaraschini: Why are we testing port `12345` separately above and then these others ? Isn't this just duplicated ?

ricardomaraschini (changes requested):

test/e2e/network_policy_enforcement.go
```diff
@@ -0,0 +1,236 @@
...
+		"app":                "openshift-controller-manager-a",
+		"controller-manager": "true",
+	}
+	routeControllerManagerLabels := map[string]string{
+		"app":                      "route-controller-manager",
+		"route-controller-manager": "true",
+	}
+
+	g.By("Creating test server pods for cross-namespace metrics checks")
+	operatorServerIPs, cleanupOperator := createServerPod(ctx, kubeClient, util.OperatorNamespace, fmt.Sprintf("np-xns-operator-%s", rand.String(5)), operatorLabels, 8443)
```
ricardomaraschini, line 106: Why don't we test against the actual metrics endpoint ? What if the operator change its labels ? Can't we end up in a scenario where we can access this "temporary server pod" but can't access the same port on the actual operator ?

test/e2e/network_policy_enforcement.go
```diff
@@ -0,0 +1,236 @@
...
+	g.By("Verifying cross-namespace access: default -> operator:8443 (any namespace can access metrics)")
+	expectConnectivity(ctx, kubeClient, "default", map[string]string{"test": "client"}, operatorServerIPs, 8443, true)
+
+	g.By("Verifying cross-namespace access: openshift-console -> controller-manager:8443")
+	expectConnectivity(ctx, kubeClient, "openshift-console", map[string]string{"app": "console"}, controllerServerIPs, 8443, true)
+}
+
+// testMetricsPortOpenOthersBlocked validates that port 8443 (metrics) is accessible
+// from various namespaces, but other ports are blocked by the default-deny policy.
+func testMetricsPortOpenOthersBlocked(ctx context.Context) {
```
ricardomaraschini: Why not merge this logic in the previous test ? I mean on `testCrossNamespaceMetricsAccess` ?  Maybe rename the function and shove these extra checks in there.

liouk:
@ricardomaraschini just to add some context to this PR based on your review -- the helpers you see in `test/e2e/network_policy.go` and `test/e2e/network_policy_enforcement.go` come from https://github.com/openshift/cluster-authentication-operator/pull/831. Due to the lack of one single repo/lib that we could place test helpers, we decided for the moment to reuse the code of that PR in all necessary components, until such a lib gets created.

test/e2e/helpers.go
kaleemsiddiqu: Removed namespaceSelectorMatches, using now nsMatch

test/e2e/helpers.go
kaleemsiddiqu: Removed ruleAllowsPort as this was used only in ingressAllowsFromNamespace which was removed as it was unused

test/e2e/helpers.go
kaleemsiddiqu: removed this along with logConnectivityBestEffortForIP

test/e2e/helpers.go
kaleemsiddiqu: removed.

test/e2e/helpers.go
kaleemsiddiqu: removed.

test/e2e/helpers.go
kaleemsiddiqu: removed.

test/e2e/helpers.go
kaleemsiddiqu: added.

test/e2e/helpers.go
kaleemsiddiqu: Removed functions restoreNetworkPolicy() and mutateAndRestoreNetworkPolicy(). All calls now use explicit timeouts with restoreNetworkPolicyWithTimeout() and mutateAndRestoreNetworkPolicyWithTimeout()

test/e2e/helpers.go
kaleemsiddiqu: it was consistent but was in hardcoded value, which replaced now with variables as you suggested which is better way.

test/e2e/helpers.go
kaleemsiddiqu: Added logging in few other cleanup functions as well e.g createConnectivityClientPod

test/e2e/network_policy.go
kaleemsiddiqu: This does not require the timeout, timeout is used for tests which are serial in nature and longer. this is for specific scenario where we expect a longer runtime

test/e2e/network_policy_enforcement.go
kaleemsiddiqu: This testing the policies along with controller manager policies as sanity and verifies that these works well along with controller manager network policies as well

test/e2e/network_policy_enforcement.go
kaleemsiddiqu: Removed this particular test and now just test with port 12345

test/e2e/network_policy_enforcement.go
kaleemsiddiqu: There was issue i think with actual pods, probably some connectivity protocol or some other issue. Now updated to test to use operator labels with test pods

test/e2e/network_policy_enforcement.go
kaleemsiddiqu: Merged testCrossNamespaceMetricsAccess and testMetricsPortOpenOthersBlocked into a single function testMetricsPortAccessControl

test/e2e/network_policy.go
kaleemsiddiqu: Ginkgo's test framework handles this timoout based on tag [Timeout:30m]

test/e2e/network_policy_enforcement.go
ricardomaraschini: This is my understanding of this test: we are testing if network policies work in general and it has nothing to do with the component in question. I am not arguing that the test has no value, I am arguing that this belongs somewhere else.

test/e2e/helpers.go
```diff
@@ -0,0 +1,895 @@
...
+
+const (
+	agnhostImage = "registry.k8s.io/e2e-test-images/agnhost:2.45"
+
+	// Poll intervals
+	defaultPollInterval = 5 * time.Second
+	fastPollInterval    = 2 * time.Second
+
+	// Timeouts
+	shortTimeout             = 30 * time.Second
```
ricardomaraschini, line 35: unused.

test/e2e/helpers.go
```diff
@@ -0,0 +1,895 @@
...
+
+// requireIngressPort validates that a policy allows a specific ingress port.
+func requireIngressPort(policy *networkingv1.NetworkPolicy, protocol corev1.Protocol, port int32) {
+	g.GinkgoHelper()
+	o.Expect(hasPortInIngress(policy.Spec.Ingress, protocol, port)).To(o.BeTrue(),
+		"%s/%s: expected ingress port %s/%d", policy.Namespace, policy.Name, protocol, port)
+}
+
+// requireIngressFromNamespace validates that a policy allows ingress from a specific namespace.
+func requireIngressFromNamespace(policy *networkingv1.NetworkPolicy, port int32, namespace string) {
```
ricardomaraschini, line 90: unused.

ricardomaraschini (changes requested):
Please check this one once more, we have 10 unused items (9 functions and 1 timeout).

ricardomaraschini:
```bash
$ gopls codeaction test/e2e/helpers.go
edit    "Delete const \"shortTimeout\"" [quickfix]
edit    "Delete function \"requireIngressFromNamespace\"" [quickfix]
edit    "Delete function \"logIngressFromNamespaceOptional\"" [quickfix]
edit    "Delete function \"requireIngressFromNamespaceOrPolicyGroup\"" [quickfix]
edit    "Delete function \"logIngressHostNetworkOrAllowAll\"" [quickfix]
edit    "Delete function \"requireEgressPort\"" [quickfix]
edit    "Delete function \"logEgressAllowAll\"" [quickfix]
edit    "Delete function \"logEgressAllowAllTCP\"" [quickfix]
edit    "Delete function \"getReadyPodIPs\"" [quickfix]
edit    "Delete function \"podMatch\"" [quickfix]
command "Browse documentation for package e2e" [source.doc]
command "Split package \"e2e\"" [source.splitPackage]
command "Show compiler optimization details for \"e2e\"" [source.toggleCompilerOptDetails]
command "Browse gopls feature documentation" [gopls.doc.features]
$
```

test/e2e/network_policy_enforcement.go
kaleemsiddiqu: These tests are doing sanity check of network policy at cluster level before proceeding to specific component checks for these. It will help us from many perspective from debugging, failure detection at cluster level which might affect component level network policies too, checking etc.

ricardomaraschini: I understand but if this test fails there is absolutely nothing we can do to fix it on this repository. More, if the feature checked by this test fails, we will already see an error  here so this is redundant.

kaleemsiddiqu: Removed.

kaleemsiddiqu:
/verified by @kaleemsiddiqu 
Executed tests successfully locally.

PR #439, opened 2026-06-08
OCPBUGS-77404: bugfix: mount /etc/containers so image controllers respect cluster-wide image registry configurations

everettraven:
# Description

Fixes [OCPBUGS-77404](https://redhat.atlassian.net/browse/OCPBUGS-77404) by updating the `openshift-controller-manager` deployment to mount the `/etc/containers` directory from the host that the image controllers using the containers/image package will read for registry configurations.

The Machine Configuration Operator is responsible for syncing cluster configurations through the image APIs to this directory on the nodes.

In order to do this, the SCC needed to be adjusted to one that allowed host path mounting. To limit the changes as much as possible I updated the SCC to `hostaccess` from `restricted-v3` so that we can mount the host path but prevent running the container as `root`, maintaining the user namespace security boundary.

## Summary by CodeRabbit

* **Chores**
  * Controller now requests the host-access security constraint; new role and rolebinding grant that access.
  * Operator synchronizes the cluster pull-secret into the controller namespace at startup.
  * Controller namespace annotated for privileged pod-security.
* **New Features**
  * Controller mounts host container configuration read-only and exposes a registry auth file via an environment variable for image pulls.
* **Tests**
  * Tests updated to expect the new registry auth env var and adjusted formatting in expected configs.

### DIFF
--- a/bindata/assets/openshift-controller-manager/deploy.yaml
+++ b/bindata/assets/openshift-controller-manager/deploy.yaml
@@ -69,6 +69,8 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.name
+        - name: REGISTRY_AUTH_FILE
+          value: /var/run/secrets/image-auth/auth.json
         livenessProbe:
           initialDelaySeconds: 30
           httpGet:
@@ -95,6 +97,8 @@ spec:
         - mountPath: /etc/containers
           name: etc-containers
           readOnly: true
+        - mountPath: /var/run/secrets/image-auth
+          name: image-auth
       volumes:
       - name: config
         configMap:
@@ -117,6 +121,12 @@ spec:
           path: /etc/containers
           type: Directory
         name: etc-containers
+      - name: image-auth
+        secret:
+          secretName: pull-secret
+          items:
+            - key: .dockercfgjson
+              path: auth.json
       nodeSelector:
         node-role.kubernetes.io/master: ""
       tolerations:
--- a/bindata/assets/openshift-controller-manager/deploy.yaml
+++ b/bindata/assets/openshift-controller-manager/deploy.yaml
@@ -39,9 +39,6 @@ spec:
         controller-manager: "true"
     spec:
       hostUsers: false
-      securityContext:
-        seccompProfile:
-          type: RuntimeDefault
       priorityClassName: system-node-critical
       serviceAccountName: openshift-controller-manager-sa
       containers:
--- a/bindata/assets/openshift-controller-manager/deploy.yaml
+++ b/bindata/assets/openshift-controller-manager/deploy.yaml
@@ -33,7 +33,7 @@ spec:
       name: openshift-controller-manager
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
-        openshift.io/required-scc: restricted-v3
+        openshift.io/required-scc: hostaccess
       labels:
         app: openshift-controller-manager-a
         controller-manager: "true"
@@ -92,6 +92,9 @@ spec:
           name: proxy-ca-bundles
         - mountPath: /tmp
           name: tmp
+        - mountPath: /etc/containers
+          name: etc-containers
+          readOnly: true
       volumes:
       - name: config
         configMap:
@@ -110,6 +113,10 @@ spec:
               path: tls-ca-bundle.pem
       - emptyDir: {}
         name: tmp
+      - hostPath:
+          path: /etc/containers
+          type: Directory
+        name: etc-containers
       nodeSelector:
         node-role.kubernetes.io/master: ""
       tolerations:
--- a/bindata/assets/openshift-controller-manager/deploy.yaml
+++ b/bindata/assets/openshift-controller-manager/deploy.yaml
@@ -122,7 +122,7 @@ spec:
         secret:
           secretName: pull-secret
           items:
-            - key: .dockercfgjson
+            - key: .dockerconfigjson
               path: auth.json
       nodeSelector:
         node-role.kubernetes.io/master: ""
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -251,6 +251,9 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 			"assets/openshift-controller-manager/networkpolicy-default-deny.yaml",
 			"assets/openshift-controller-manager/route-controller-manager-networkpolicy-allow.yaml",
 			"assets/openshift-controller-manager/route-controller-manager-networkpolicy-default-deny.yaml",
+
+			"assets/openshift-controller-manager/scc-role.yaml",
+			"assets/openshift-controller-manager/scc-rolebinding.yaml",
 		},
 		resourceapply.NewKubeClientHolder(kubeClient),
 		opClient,
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -131,6 +131,14 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		controllerConfig.EventRecorder,
 	)
 
+	err = resourceSyncer.SyncSecret(
+		resourcesynccontroller.ResourceLocation{Namespace: "openshift-config", Name: "pull-secret"},
+		resourcesynccontroller.ResourceLocation{Namespace: util.TargetNamespace, Name: "pull-secret"},
+	)
+	if err != nil {
+		return fmt.Errorf("configuring global pull-secret syncing: %w", err)
+	}
+
 	if !cache.WaitForCacheSync(ctx.Done(), configInformers.Config().V1().ClusterVersions().Informer().HasSynced) {
 		klog.Errorf("timed out waiting for configInformers ClusterVersions")
 		return fmt.Errorf("timed out waiting for configInformers ClusterVersions")
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -132,8 +132,8 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 	)
 
 	err = resourceSyncer.SyncSecret(
-		resourcesynccontroller.ResourceLocation{Namespace: "openshift-config", Name: "pull-secret"},
 		resourcesynccontroller.ResourceLocation{Namespace: util.TargetNamespace, Name: "pull-secret"},
+		resourcesynccontroller.ResourceLocation{Namespace: "openshift-config", Name: "pull-secret"},
 	)
 	if err != nil {
 		return fmt.Errorf("configuring global pull-secret syncing: %w", err)
--- a/bindata/assets/openshift-controller-manager/ns.yaml
+++ b/bindata/assets/openshift-controller-manager/ns.yaml
@@ -8,3 +8,6 @@ metadata:
   labels:
     openshift.io/cluster-monitoring: "true"
     openshift.io/run-level: "" # specify no run-level turns it off on install and upgrades
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/warn: privileged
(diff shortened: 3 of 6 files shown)


bindata/assets/openshift-controller-manager/deploy.yaml
```diff
@@ -110,6 +113,10 @@ spec:
               path: tls-ca-bundle.pem
       - emptyDir: {}
         name: tmp
+      - hostPath:
```
everettraven, line 117: Open Question: Does this automatically handle pull secrets for private registries as well?

everettraven: It doesn't, but I've opened https://github.com/openshift/openshift-controller-manager/pull/447 to help with this.

bindata/assets/openshift-controller-manager/deploy.yaml
ehearne-redhat: I think we need to remove this as `e2e-upgrade` reports 

```bash
event happened 36 times, something is wrong: namespace/openshift-controller-manager hmsg/827f4eb8d4 replicaset/controller-manager-99bbc9cc8 - reason/FailedCreate Error creating: pods "controller-manager-99bbc9cc8-" is forbidden: unable to validate against any security context constraint: [metadata.annotations[seccomp.security.alpha.kubernetes.io/pod]: Forbidden: seccomp may not be set, metadata.annotations[container.seccomp.security.alpha.kubernetes.io/controller-manager]: Forbidden: seccomp may not be set] (21:59:34Z) result=reject }
```

ehearne-redhat:
@bpalmer The changes in general look fine to me, the seccomp concern is the only one I have. Could you take a look at that and let me know if this is an issue? I dont see any mention https://docs.redhat.com/en/documentation/openshift_container_platform/4.22/html/authentication_and_authorization/managing-pod-security-policies#default-sccs_configuring-internal-oauth about it not being allowed.

However it says privileged containers cannot use `seccompprofile` at https://docs.redhat.com/en/documentation/openshift_container_platform/4.22/html/security_and_compliance/seccomp-profiles .

everettraven:
This last failure has a really odd failure. It says the `openshift-controller-manager-sa` service account doesn't exist but I can see an event from when it was created and the audit logs show no sign of that service account being deleted (or created).

Going to see if maybe this was a fluke.

/retest-required

ehearne-redhat:
@everettraven - do we need to add `pod-security.kubernetes.io` annotations to https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/bindata/assets/openshift-controller-manager/ns.yaml as the pod is now privileged ? 

It says [here](https://docs.redhat.com/en/documentation/openshift_container_platform/4.22/html/authentication_and_authorization/understanding-and-managing-pod-security-admission#security-context-constraints-psa-about_understanding-and-managing-pod-security-admission) that pods that

>Pods that do not comply with the pod security admission defined globally or at the namespace level are not admitted to the cluster and cannot run.


It says [here](https://docs.redhat.com/en/documentation/openshift_container_platform/4.22/html/authentication_and_authorization/understanding-and-managing-pod-security-admission#permanently-disabled-namespaces ) that

>All system-created namespaces that are prefixed with openshift- , except for openshift-operators

have pod security admission synchronization disabled permanently.

Looking at the must-gather, I can see pod security admission is set to `restricted`. 

```yaml
spec:
  logLevel: Normal
  managementState: Managed
  observedConfig:
    admission:
      pluginConfig:
        PodSecurity:
          configuration:
            defaults:
              audit: restricted
              audit-version: latest
              enforce: restricted
              enforce-version: latest
              warn: restricted
              warn-version: latest
```

[Kubernetes docs](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted) states under this configuration, we need `RuntimeDefault` seccomp profile. Since we are running in `hostaccess` SCC now, we cannot use this profile. 

Looking at [the same doc](https://kubernetes.io/docs/concepts/security/pod-security-standards/#privileged), the minimum we can set `pod-security.kubernetes.io` annotations to is `privileged` because of the `hostaccess` SCC.

So, we will probably need to change the namespace to use these labels:


```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: openshift-controller-manager
  annotations:
    openshift.io/node-selector: ""
    workload.openshift.io/allowed: "management"
  labels:
    openshift.io/cluster-monitoring: "true"
    openshift.io/run-level: "" # specify no run-level turns it off on install and upgrades
    pod-security.kubernetes.io/enforce: privileged
    pod-security.kubernetes.io/audit: privileged
    pod-security.kubernetes.io/warn: privileged
```

I have found an example of an `openshift-*` namespace using these labels [here](https://github.com/openshift/cluster-openshift-apiserver-operator/blob/98775a10673342e09ed518a29cae15288624a522/bindata/v3.11.0/openshift-apiserver/ns.yaml#L13).

everettraven:
@ehearne-redhat Yeah, we will need to do that but I don't think that is what is causing the failures here. I'm trying to sort out why we are seeing an issue with pod creation related to non-existence of the service account. Something seems to be removing the service account before we attempt to create the pods from the deployment here.

pkg/operator/starter.go
ehearne-redhat, line 315: @everettraven I believe there is a race condition as to when the manifests are applied. They appear to be applied concurrently, not sequentially, as the deployment and SA/SCCs/etc are managed by two separate controllers.

We probably need to add a check somewhere near https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go#L134 to check if the service account + related resources are present before applying the deployment manifest. 

Adding something like...
```golang
// Verify service account exists before creating deployment
  _, err = c.kubeClient.CoreV1().ServiceAccounts(util.TargetNamespace).Get(context.TODO(),
  "openshift-controller-manager-sa", metav1.GetOptions{})
  if err != nil {
      ocmErrors = append(ocmErrors, fmt.Errorf("%q: waiting for service account: %w", operandName, err))
      return syncReturn(c, ocmErrors, rcmErrors, originalOperatorConfig, operatorConfig)
  }

// repeat the same for scc + rolebinding too. 
```
should ensure we don't hit the race condition.

everettraven: I don't think a race condition is the problem we are seeing.

I'm seeing an event propagate that says the `openshift-controller-manager-sa` service account is getting created but there appears to be no evidence of it existing in the must gather that is collected from the test clusters.

Even if the controller enters a degraded state for some amount of time, eventual consistency should ensure that _eventually_ the service account we are waiting for exists and the deployment will continue to roll out.

ehearne-redhat: Usually when I check must gather I never see serviceaccounts.yaml or something like that in most namespaces. Only some namespaces I have seen this file. 

I see this in `/core/events.yaml`

```bash
 message: 'Error creating: pods "controller-manager-f4dc56774-" is forbidden: error
    looking up service account openshift-controller-manager/openshift-controller-manager-sa:
    serviceaccount "openshift-controller-manager-sa" not found'
```

```bash
message: 'Error creating: pods "controller-manager-f4dc56774-" is forbidden: unable
    to validate against any security context constraint: provider "hostaccess": Forbidden:
    not usable by user or serviceaccount'
```

but then it follows up with

```bash
message: 'Created pod: controller-manager-f4dc56774-xbj7h'
```

There are also messages saying it can't find the `config` configMap, `serving-cert` secret too.

And also 

```bash
message: 'MountVolume.SetUp failed for volume "image-auth" : references non-existent
    secret key: .dockercfgjson'
```

ehearne-redhat: What's interesting is the 19s difference in timestamp between pod creation and deletion.

```bash
lastTimestamp: "2026-06-12T13:37:31Z"
  message: 'Created pod: controller-manager-f4dc56774-xbj7h'
```

```bash
lastTimestamp: "2026-06-12T13:37:50Z"
  message: 'Deleted pod: controller-manager-f4dc56774-xbj7h'
```

bindata/assets/openshift-controller-manager/deploy.yaml
```diff
@@ -110,6 +114,16 @@ spec:
...
+          path: /etc/containers
+          type: Directory
+        name: etc-containers
+      - name: image-auth
+        secret:
+          secretName: pull-secret
+          items:
+            - key: .dockercfgjson
```
ehearne-redhat: should this be `.dockerconfigjson` ?

https://github.com/openshift/openshift-controller-manager/blob/master/vendor/k8s.io/api/core/v1/types.go#L7946-L7953

pkg/operator/starter.go
ehearne-redhat: I'm thinking it wouldn't hurt to add checks to make sure these resources exist before we apply the `deploy.yaml` manifest. The worst thing that could happen is that we discover the controller works fine. :D 

For me, the logs show that manifests are missing which cause errors. If we can make sure the manifests are there first before applying, we can at least rule out that scenario. We could be looking at different logs however... :)

everettraven: Ah, I think you found the smoking gun with those events. I hadn't yet found time to dig deeper into the events timeline yet so thanks for that. The wrong key reference definitely could cause this.

Also TIL that the service accounts don't seem to show up in the must gathers :). I wonder why that is?

ehearne-redhat: I spent a little time investigating why this is, and it actually comes down to not even using the must-gather command at all!

So, the interesting thing is, service accounts are captured in some namespaces.

```bash
ehearne-mac:quay-proxy-ci-openshift-org-openshift-ci-sha256-9da0e7fd8e430ccec3c8da980b0718e0f826c599f580a98189679e9c584370f1 ehearne$ find . -path "*/serviceaccounts/*"
./namespaces/openshift-cluster-olm-operator/core/serviceaccounts/cluster-olm-operator.yaml
./namespaces/openshift-cloud-network-config-controller/core/serviceaccounts/cloud-network-config-controller.yaml
./namespaces/openshift-operator-controller/core/serviceaccounts/operator-controller-controller-manager.yaml
./namespaces/openshift-cluster-csi-drivers/core/serviceaccounts/aws-ebs-csi-driver-operator.yaml
./namespaces/openshift-multus/core/serviceaccounts/multus.yaml
./namespaces/openshift-multus/core/serviceaccounts/multus-ancillary-tools.yaml
./namespaces/openshift-multus/core/serviceaccounts/metrics-daemon-sa.yaml
./namespaces/openshift-multus/core/serviceaccounts/multus-ac.yaml
./namespaces/openshift-multus/core/serviceaccounts/multus-node-identity.yaml
./namespaces/openshift-network-operator/core/serviceaccounts/iptables-alerter.yaml
./namespaces/openshift-network-console/core/serviceaccounts/networking-console-plugin.yaml
./namespaces/openshift-network-node-identity/core/serviceaccounts/network-node-identity.yaml
./namespaces/openshift-cloud-credential-operator/core/serviceaccounts/pod-identity-webhook.yaml
./namespaces/openshift-cluster-storage-operator/core/serviceaccounts/csi-snapshot-controller-operator.yaml
./namespaces/openshift-insights/core/serviceaccounts/gather.yaml
./namespaces/openshift-insights/core/serviceaccounts/operator.yaml
./namespaces/openshift-ovn-kubernetes/core/serviceaccounts/ovn-kubernetes-control-plane.yaml
./namespaces/openshift-ovn-kubernetes/core/serviceaccounts/ovn-kubernetes-node.yaml
./namespaces/openshift-network-diagnostics/core/serviceaccounts/network-check-target.yaml
./namespaces/openshift-network-diagnostics/core/serviceaccounts/network-diagnostics.yaml
ehearne-mac:quay-proxy-ci-openshift-org-openshift-ci-sha256-9da0e7fd8e430ccec3c8da980b0718e0f826c599f580a98189679e9c584370f1 ehearne$ 
```

It comes down to what namespaces use `oc adm inspect --namespace` as `oc adm inspect` when ns is specified will gather all resources into the dump, including service accounts. 

https://github.com/openshift/oc/blob/40ce70fca070aafb0273563ce5a7f0a5ba1fcdb2/pkg/cli/admin/inspect/namespace.go#L22

Any namespace that has `update-service-operator` will get serviceaccounts included to inspect in must gather as a result. 

https://github.com/openshift/must-gather/blob/16ac27eedcc79f5b57d77eb01e6b187b84fe7daa/collection-scripts/gather_osus#L9

Funnily enough, this was included as part of the first PR for must-gather :) https://github.com/openshift/must-gather/pull/1 .

